PR #122, opened 2023-07-20
Profile and Team Cards, Profile Buttons

### LINKED ISSUE #109: Create ProfileCard Component
image

- Create the Profile Card Component
- Only need to work on frontend, do not need to implement middle end

### LINKED ISSUE #110: Create Team Card Component
image

- Create Team Card Component
- Only need to work on frontend do not need to implement middle end

### LINKED ISSUE #111: Create ProfileButton Component
image
image

- Make props for the text, icon, and color scheme
- Checkout https://github.com/acm-ucr/hackathon-website/blob/dev/src/components/Tag.jsx#L3 for reference
- Create a `ProfileButton.jsx` component

rfairooz:
Hi so right now, I have the profile buttons rendered based on the name of the button that is passed in. That is the only prop. I thought it was easier to have one file deal with the styling while the other files can just call whichever button they want - like the Tag.jsx. Let me know if that is ok. If not, I can change the buttons to just have props (name, color etc) and we can specify when we call. 

Screenshot 2023-07-20 at 11 37 02 AM
Screenshot 2023-07-20 at 11 37 48 AM

### DIFF
--- a/src/components/ProfileButton.jsx
+++ b/src/components/ProfileButton.jsx
@@ -1,0 +1,31 @@
+import React from "react";
+import { BiEdit } from "react-icons/bi";
+import { GiCancel } from "react-icons/gi";
+
+const buttonType = {
+  edit: {
+    background: "bg-hackathon-blue-100",
+    text: "text-hackathon-gray",
+    icon: <BiEdit />,
+  },
+
+  cancel: {
+    background: "bg-hackathon-gray",
+    text: "text-hackathon-blue-200",
+    icon: <GiCancel />,
+  },
+};
+
+const ProfileButton = ({ name }) => {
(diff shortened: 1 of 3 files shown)


shahdivyank:
Pass in all the attributes as props. For colors, you can simplify it to something like "gray" or "blue" and then keep a map of what those colors actually are. Checkout the tag component for a similar implementation

rfairooz:
> Pass in all the attributes as props. For colors, you can simplify it to something like "gray" or "blue" and then keep a map of what those colors actually are. Checkout the tag component for a similar implementation

Makes sense, fixed the props. Renders the same as the screenshots.

shahdivyank (approved):

PR #147, opened 2023-07-27
Agent einstein/fix hover bug

### LINKED ISSUE #95: BUG: Fix Hover Bug
image

- When it is a status, a user should not be able to see this hover behavior, this implies there is an action that can be completed
- Remove it for the status, but ensure it works for the action buttons in the toolbar

AgentEinstein:
image
Can't demonstrate this because of how screenshots work but I promise this is fixed


SafeDuck:
run "npm run format"
commit again

PR #168, opened 2023-08-02
created Label.jsx and added it to stats page

### LINKED ISSUE #86: Create Label.jsx
image
image

- Create a new Component called `Label` and add a prop for the `text`
- Call it within the statistics page

aditithanekar:
The "status" part is the label
image

### DIFF
--- a/src/components/Label.jsx
+++ b/src/components/Label.jsx
@@ -1,0 +1,11 @@
+import React from "react";
+
+const Label = ({ label }) => {
+  return (
+    <div className="flex justify-content-center p-1 px-4 bg-hackathon-gray text-secondary rounded-pill h-50">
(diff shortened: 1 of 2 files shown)


src/components/Label.jsx
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+
+const Label = ({ label }) => {
+  return (
+    <div className="flex justify-content-center p-1 px-4 bg-hackathon-gray text-secondary rounded-pill h-50">
```
shahdivyank: don't hard code the height, use padding to make it responsive instead

shahdivyank (changes requested):

shahdivyank (approved):

PR #182, opened 2023-08-03
Adding live timer for check-in page

### LINKED ISSUE #112: Change Text
image

- Change it to show the current time with hours, minutes, seconds
- This prevents people from screenshotting the page and using it multiple times

torisutansan02:
24 hour timer for the user check-in page.


src/components/CheckIn.jsx
```diff
@@ -9,6 +11,23 @@ const mockuser = {
   email: "big.chungus@email.com",
   pfp: LOGO,
 };
+
+const Timer = () => {
+  const date = new Date();
+  const [currentTime, setCurrentTime] = React.useState("");
```
shahdivyank: import useState directly

src/components/CheckIn.jsx
```diff
@@ -9,6 +11,23 @@ const mockuser = {
   email: "big.chungus@email.com",
   pfp: LOGO,
 };
+
+const Timer = () => {
+  const date = new Date();
```
shahdivyank: use this value directly within the useState. You can simplify the code to just use the useState and eliminate the need for the second variable

shahdivyank (changes requested):

src/components/CheckIn.jsx
```diff
@@ -40,7 +39,6 @@ const CheckinPage = () => {
           <Image
             src={mockuser.pfp}
             className="w-40 h-40 rounded-full overflow-hidden"
-            alt="Picture of user's profile"
```
shahdivyank: alt should be there

src/components/CheckIn.jsx
```diff
@@ -10,7 +10,6 @@ const mockuser = {
   name: "Big Chungus",
   email: "big.chungus@email.com",
   pfp: LOGO,
-  uid: 123456789,
```
shahdivyank, line 11: uid should also be there

shahdivyank (changes requested):

shahdivyank (approved):

PR #201, opened 2023-08-05
3x3 piechart

### LINKED ISSUE #179: Add 9 PieCharts
3 x 3 Pie chart configuration

SafeDuck:
image


src/app/admin/statistics/page.js
```diff
@@ -5,16 +5,28 @@ import PieChart from "@/components/PieChart";
 import ProtectedPage from "@/components/ProtectedPage";
 
 const StatisticsPage = () => {
```
shahdivyank, line 7: Don't make a separate function for render rows. Integrate it directly. Us bootstrap grid to make it a single call instead of a nested call. U can do 1 row with multiple col where each col is a chart and then u can specify the size

shahdivyank (changes requested):

shahdivyank (approved):

PR #268, opened 2023-08-15
download csv file middle-end

### LINKED ISSUE #193: Download Tables
https://sheetjs.com

- Create a utility function to download data 
- Store this utility in src/utils

SafeDuck:
image


src/components/Admin/Judges.jsx
```diff
@@ -135,6 +135,8 @@ const Judges = () => {
         objects={filteredJudges}
         filters={filters}
         reset={judges}
+        download={judges}
+        fileName="judges"
```
shahdivyank: Can you add a timestamp to the filenames, this is to prevent having judges-1.csv and such being downloaded. If we were to click download multiple times, ideally it should be timestamped so we know the most recent version. You only need to include day + hour + minute (might remove minute but lets add it for now).

shahdivyank: also it should include the hackathon name, since one person could be in charge of multiple hackathons, we need a way to differentiate between hackathon

src/components/Admin/Toolbar.jsx
```diff
@@ -13,6 +14,8 @@ const Toolbar = ({
   objects,
   filters,
   reset,
+  download,
+  fileName,
```
shahdivyank: call the prop: `file`

src/components/Admin/Toolbar.jsx
```diff
@@ -72,7 +75,58 @@ const Toolbar = ({
       );
     }
   };
+  const csvData = objects.map((download) => {
+    const rowData = {};
 
+    if (download.name) {
```
shahdivyank: Avoid the if statements, this needs to be more abstract as we may add or delete columns so it needs to work out, for now map through the objects entries and use those values directly. Later we need to add an option to delete/hide particular columns (this is a later issue)

src/components/Admin/Toolbar.jsx
```diff
@@ -110,9 +164,13 @@ const Toolbar = ({
           Reset
         </button>
       </div>
-      <button>
+      <CSVLink
+        data={csvData}
+        filename={`${fileName}.csv`}
```
shahdivyank: put the .csv when u initially pass it in at the page level, so like Mentors.jsx and such

shahdivyank (changes requested):

src/components/Admin/Toolbar.jsx
SafeDuck: This will be done by the end of the day

src/components/Admin/Toolbar.jsx
```diff
@@ -62,7 +64,36 @@ const Toolbar = ({
     );
     setToggle(!toggle);
   };
+  // const csvData = objects.entries(download).map(entry => rowData[entry[0]] = entry[1]);
```
shahdivyank: remove commented out code

src/components/Admin/Judges.jsx
```diff
@@ -137,6 +137,7 @@ const Judges = () => {
         objects={filteredJudges}
         filters={filters}
         reset={judges}
+        file="JUDGES.csv"
```
shahdivyank: apologies, but now that we r doing the more complex naming scheme, u can keep the .csv at the final stage with the env variables and stuff. just need JUDGES here.

src/components/Admin/Toolbar.jsx
```diff
@@ -62,7 +64,36 @@ const Toolbar = ({
...
+  // const csvData = objects;
+  const csvData = objects.map((download) => {
+    const rowData = {};
+
+    Object.entries(download).forEach(([key, value]) => {
+      if (key === "members" && Array.isArray(value) && value.length > 0) {
```
shahdivyank: don't make the assumption that it will be members always, we could have other fields that are arrays and it would work for that as well. u can simply check for the array type and length and then use the join functions

src/components/Admin/Toolbar.jsx
```diff
@@ -62,7 +64,36 @@ const Toolbar = ({
...
+        const memberNames = value.map((member) => member.name).join(", ");
+        const memberEmails = value.map((member) => member.email).join(", ");
+
+        rowData.members = `Names: ${memberNames}, Emails: ${memberEmails}`;
+      } else {
+        if (typeof value === "boolean") {
+          rowData[key] = value ? "Yes" : "No";
```
shahdivyank: keep it true and false instead of yes and no

src/components/Admin/Toolbar.jsx
```diff
@@ -62,7 +64,36 @@ const Toolbar = ({
...
+        } else {
+          rowData[key] = value || "";
```
shahdivyank: in what situations does this occur?

src/components/Admin/Toolbar.jsx
```diff
@@ -62,7 +64,36 @@ const Toolbar = ({
...
+  // const csvData = objects;
+  const csvData = objects.map((download) => {
+    const rowData = {};
+
+    Object.entries(download).forEach(([key, value]) => {
+      if (key === "members" && Array.isArray(value) && value.length > 0) {
+        const memberNames = value.map((member) => member.name).join(", ");
+        const memberEmails = value.map((member) => member.email).join(", ");
+
+        rowData.members = `Names: ${memberNames}, Emails: ${memberEmails}`;
```
shahdivyank: try not to hard code the names and emails, try to use the keys for it

src/components/Admin/Toolbar.jsx
```diff
@@ -62,7 +64,36 @@ const Toolbar = ({
...
+          rowData[key] = value || "";
+        }
+      }
+    });
+
+    return rowData;
+  });
 
+  const currentDate = new Date();
+  const formattedDate = currentDate
```
shahdivyank: I have a feeling there is a way to combine the 2 statements into one format

src/components/Admin/Toolbar.jsx
```diff
@@ -103,12 +134,13 @@ const Toolbar = ({
...
         </button>
       </div>
-      <button>
-        <FaDownload
-          size={22.5}
-          className="ml-4 text-hackathon-darkgray hover:opacity-70 duration-150"
-        />
-      </button>
+      <CSVLink
+        data={csvData}
```
shahdivyank: call csvData just data

src/components/Admin/Toolbar.jsx
SafeDuck: image
bool is needed

SafeDuck: otherwise it is blank

SafeDuck: the last line is needed. Here is what it looks like when you remove both:
image

src/components/Admin/Participants.jsx
```diff
@@ -35,6 +35,8 @@ const Participants = () => {
         setObjects={setParticipants}
         objects={participants}
         filters={filters}
+        reset={participants}
```
shahdivyank: what do you need the reset for?

src/components/Admin/Toolbar.jsx
```diff
@@ -4,9 +4,34 @@ import Checkbox from "../Checkbox";
 import { HiSearch } from "react-icons/hi";
 import Tag from "./Tag.jsx";
 import { FaDownload, FaTrashAlt } from "react-icons/fa";
+import { CSVLink } from "react-csv";
 import { tagColor } from "@/data/Tags";
 
-const Toolbar = ({ input, setInput, tags, setObjects, objects, filters }) => {
+function convert(input) {
```
shahdivyank: arrow functions

src/components/Admin/Toolbar.jsx
```diff
@@ -71,6 +96,20 @@ const Toolbar = ({ input, setInput, tags, setObjects, objects, filters }) => {
     setToggle(!toggle);
   };
 
+  const data = objects.map((items) => {
```
shahdivyank: this should be in a function, this is going to run every time react rerenders

src/components/Admin/Toolbar.jsx
```diff
@@ -71,6 +96,20 @@ const Toolbar = ({ input, setInput, tags, setObjects, objects, filters }) => {
...
+    const rowData = {};
+    Object.entries(items).map(
+      ([key, value]) => (rowData[key] = convert(value))
+    );
+    return rowData;
+  });
+
+  const formattedDate = new Date()
```
shahdivyank: should be in a function

SafeDuck:
Hide the uid, links, dropdown, etc. Needs review now to get pro merged to main

shahdivyank (changes requested):

src/components/Admin/Toolbar.jsx
```diff
@@ -4,9 +4,34 @@ import Checkbox from "../Checkbox";
...
 import { FaDownload, FaTrashAlt } from "react-icons/fa";
+import { CSVLink } from "react-csv";
 import { colors } from "@/data/Tags";
 
-const Toolbar = ({ input, setInput, tags, setObjects, objects, filters }) => {
+function convert(input) {
+  if (Array.isArray(input)) {
+    return input.map(convert).join(", ");
+  } else if (typeof input === "boolean") {
+    return input ? "True" : "False";
```
shahdivyank: what's the default return? like just return input?

src/components/Admin/Toolbar.jsx
```diff
@@ -4,9 +4,34 @@ import Checkbox from "../Checkbox";
...
+  } else if (typeof input === "object") {
+    return Object.entries(input)
```
shahdivyank: can u show an example of this mapping and join, like what does it look like?

src/components/Admin/Toolbar.jsx
```diff
@@ -70,9 +95,25 @@ const Toolbar = ({ input, setInput, tags, setObjects, objects, filters }) => {
     );
     setToggle(!toggle);
   };
+  const blacklistArray = ["uid", "selected", "hidden", "links", "dropdown"];
```
shahdivyank: just call it blacklist

src/components/Admin/Toolbar.jsx
```diff
@@ -70,9 +95,25 @@ const Toolbar = ({ input, setInput, tags, setObjects, objects, filters }) => {
...
+    });
+    return rowData;
+  });
   const handleDelete = () => {
     setObjects(objects.filter((object) => !object.selected));
   };
+  const formattedDate = new Date()
+    .toLocaleDateString("en-US")
+    .replace(/ /g, "_");
+  const formattedTime = new Date()
```
shahdivyank: should be a way to combine the 2 formatted date classes

src/components/Admin/Toolbar.jsx
SafeDuck: image

yes, it's on line 22

src/components/Admin/Toolbar.jsx
SafeDuck: it turns {a: 1, b: 2} to "a: 1, b: 2"

shahdivyank: - can we just do 1, 2, 3...
- does the generated CSV have a table header?

src/components/Admin/Judges.jsx
```diff
@@ -34,6 +34,8 @@ const Judges = () => {
         setObjects={setJudges}
         objects={judges}
         filters={filters}
+        reset={judges}
```
shahdivyank: remove reset

src/components/Admin/Toolbar.jsx
shahdivyank: what happens when u just write `return input` in this situation?

src/components/Admin/Toolbar.jsx
```diff
@@ -4,9 +4,34 @@ import Checkbox from "../Checkbox";
...
+  if (Array.isArray(input)) {
+    return input.map(convert).join(", ");
+  } else if (typeof input === "boolean") {
+    return input ? "True" : "False";
+  } else if (typeof input === "object") {
+    return Object.entries(input)
+      .map(([key, value]) => `${key}: ${convert(value)}`)
+      .join(", ");
+  } else if (typeof input === "number") {
+    return input.toString();
```
shahdivyank: do we need to convert to a string? does it not work when we have a number?

src/components/Admin/Toolbar.jsx
SafeDuck: image
image

src/components/Admin/Toolbar.jsx
SafeDuck: - I don't know about 1,2,3...
- There is no table header prop that you can pass that I know of. We would need to do some hard coding of the "data" prop, so then we get a "header" in the CSV file

src/components/Admin/Toolbar.jsx
SafeDuck: It was just in case we needed it. According to my comment above with the screenshots, it seems like we don't need it

src/components/Admin/Toolbar.jsx
shahdivyank: I mean just for boolean types, if you just return input on boolean types, what is the value?

src/components/Admin/Toolbar.jsx
shahdivyank: headers is getting passed into the table component and the toolbar component, so we should be able to add the headers as the first row

src/components/Admin/Toolbar.jsx
shahdivyank: remove the toString then, if we dont need it

src/components/Admin/Toolbar.jsx
SafeDuck: tested it, we don't need it

SafeDuck: just return input

src/components/Admin/Toolbar.jsx
```diff
@@ -4,10 +4,30 @@ import Checkbox from "../Checkbox";
...
 import Tag from "./Tag.jsx";
 import { FaDownload, FaTrashAlt } from "react-icons/fa";
+import { CSVLink } from "react-csv";
 import { colors } from "@/data/Tags";
 import Popup from "./Popup";
 
-const Toolbar = ({ input, setInput, tags, setObjects, objects, filters }) => {
+const convert = (input) => {
+  if (Array.isArray(input)) {
+    return input.map(convert).join(", ");
```
shahdivyank: dont need the map

src/components/Admin/Toolbar.jsx
```diff
@@ -4,10 +4,30 @@ import Checkbox from "../Checkbox";
...
+  } else if (typeof input === "object") {
+    return Object.values(input).map(convert).join(", ");
```
shahdivyank: dont need the map

src/components/Admin/Toolbar.jsx
```diff
@@ -4,10 +4,30 @@ import Checkbox from "../Checkbox";
...
+  } else {
```
shahdivyank: remove the else and just have the return

src/components/Admin/Toolbar.jsx
```diff
@@ -72,9 +92,35 @@ const Toolbar = ({ input, setInput, tags, setObjects, objects, filters }) => {
     );
     setToggle(!toggle);
   };
+  const blacklist = ["uid", "selected", "hidden", "links", "dropdown"];
+  const mapObjectsToCSVData = (objects, blacklist, headers) => {
+    const data = [];
+    const columnNames = headers
+      .map((header) => header.text)
```
shahdivyank: u should be able to combine the map and filter and do it in one go

src/components/Admin/Toolbar.jsx
```diff
@@ -72,9 +92,35 @@ const Toolbar = ({ input, setInput, tags, setObjects, objects, filters }) => {
...
+      .filter((key) => key && !blacklist.includes(key));
+    data.push(columnNames);
+    objects.forEach((item) => {
+      const rowData = {};
+      for (const key of columnNames) {
+        const value = item[key];
+        rowData[key] = convert(value);
+      }
+      data.push(Object.values(rowData));
```
shahdivyank: if u just need an array, then just store it as an array originally, it should remove the need for Object.values

src/components/Admin/Toolbar.jsx
```diff
@@ -72,9 +92,35 @@ const Toolbar = ({ input, setInput, tags, setObjects, objects, filters }) => {
...
+    return data;
+  };
+
+  const data = mapObjectsToCSVData(objects, blacklist, headers);
   const handleDelete = () => {
     setObjects(objects.filter((object) => !object.selected));
   };
+  const date = new Date();
+  const formattedDate = date.toLocaleDateString("en-US").replace(/ /g, "_");
+  const formattedTime = date
```
shahdivyank: see if this date format time can be converted with date format time

src/components/Admin/Toolbar.jsx
```diff
@@ -4,10 +4,28 @@ import Checkbox from "../Checkbox";
...
-const Toolbar = ({ input, setInput, tags, setObjects, objects, filters }) => {
+const convert = (input) => {
+  if (Array.isArray(input)) {
+    return input.join(", ");
+  } else if (typeof input === "object") {
+    return Object.values(input).join(", ");
+  }
+  return input;
+};
+const Toolbar = ({
```
shahdivyank, line 20:
```suggestion

const Toolbar = ({
```

src/components/Admin/Toolbar.jsx
```diff
@@ -72,9 +91,41 @@ const Toolbar = ({ input, setInput, tags, setObjects, objects, filters }) => {
...
+    const columns = headers
+      .filter((header) => header.text && !blacklist.includes(header.text))
+      .map((header) => header.text);
+    // divyank's code
+    // const columns = headers.filter((header) => {
+    //   if (header.text && !blacklist.includes(header.text)) {
+    //     return header.text;
+    //   }
+    // });
+    const data = [columns];
```
shahdivyank, line 106: why is this needed?

src/components/Admin/Toolbar.jsx
```diff
@@ -72,9 +91,41 @@ const Toolbar = ({ input, setInput, tags, setObjects, objects, filters }) => {
...
   };
+  const date = new Date();
+  const formattedDateTime = date
+    .toLocaleString("en-US", {
+      month: "2-digit",
+      day: "2-digit",
+      year: "numeric",
+      hour: "2-digit",
+      minute: "2-digit",
+      hour12: true,
```
shahdivyank: keep it 24 hour, also make sure its UTC time

src/components/Admin/Toolbar.jsx
SafeDuck: It's the same thing as data.push(columns)
It's adding the headers to the top of the csv file

shahdivyank (approved):

menthy-wu (approved):

Issue #205, opened 2023-08-05
Add Admin Page


shahdivyank:
![](https://user-images.githubusercontent.com/36285610/258561373-d749fe91-04d0-4e18-9d46-adf84f82d492.png)

- Add an Admin tab with the FaLock icon
- Make sure to create a page for it, so it goes to /admin
- Remove the auto redirect that takes /admin to /admin/participants

PR #271, opened 2023-08-16
Seanquiambao/admin page

### LINKED ISSUE #206: Code Admin Page
image

seanquiambao:
The page is pretty much finished but the page.js is in the wrong directory, so the URL is in /admin/admin. Not sure if I should put it in the appropriate directory now or after issue #205 gets resolved. Other than that, let me know if there's anything I need to change!

adminpage

- Closes issue #206


src/components/Admin/Tag.jsx
```diff
@@ -40,6 +40,22 @@ const colors = {
     background: "bg-hackathon-judge-student-bg",
     text: "text-hackathon-judge-student-text",
   },
+  marketing: {
```
shahdivyank: make these color based not role based

src/components/Admin/Admins.jsx
```diff
@@ -0,0 +1,168 @@
...
+    affiliation: "off",
+    status: "off",
+  });
+
+  const tags = [
+    {
+      color: "green",
+      text: "accept",
+      name: "Accept",
+      onClick: () => {
```
shahdivyank: these should be populated with the proper functions, checkout the judges page for code

src/components/Admin/Admins.jsx
```diff
@@ -0,0 +1,168 @@
...
+      },
+    },
+    {
+      color: "red",
+      text: "remove",
+      name: "Remove",
+      onClick: (setToggle) => {
+        setToggle(false);
+        setFilteredAdmins(
+          filteredAdmins.filter((a) => {
```
shahdivyank, line 103: this should only be updating status, not removing them

src/components/Admin/Admins.jsx
```diff
@@ -0,0 +1,168 @@
...
+              {header.name}
+              {header.icon && (
+                <SortIcon
+                  name={header.name.toLowerCase()}
+                  sorts={sorts}
+                  setSorts={setSorts}
+                  setfilteredObjects={setFilteredAdmins}
+                  objects={filteredAdmins}
+                  reset={{
+                    name: "off",
```
shahdivyank: this reset doesn't match with the sorts u have at the top. Make a variable that stores this reset and pass it into the use state as well as here. check the other pages for reference

src/components/Admin/Toolbar.jsx
```diff
@@ -44,7 +44,10 @@ const Toolbar = ({
         let boolean = false;
 
         Object.keys(filters).map((value) => {
-          if (a.status === value && filters[value]) {
+          if (
```
shahdivyank: remove the affiliation. for the filters, we will be changing them to pending, approved, rejected as the other pages as well

src/components/Admin/Filters.jsx
```diff
@@ -19,7 +19,7 @@ const Filters = ({
 
         Object.keys(filterValues).map((value) => {
           if (
-            a.status === value &&
+            (a.status === value || a.affiliation === value) &&
```
shahdivyank: remove affiliation, filtering will only work with status

shahdivyank (changes requested):

src/components/Admin/Tag.jsx
seanquiambao: not sure if this is what you meant, but i changed the admin.jsx tag to use the judges colors instead.

src/app/admin/admin/page.js
```diff
@@ -0,0 +1,16 @@
+import React from "react";
+import ProtectedPage from "@/components/ProtectedPage";
+import Admins from "@/components/Admin/Admins";
+
+const AdminPage = () => {
+  return (
+    <ProtectedPage title="Admin" restrictions={["admin"]}>
+      <div className="w-11/12">
+        <title>Admin</title>
```
shahdivyank: u don't need this, this is integrated alr within protected pages

shahdivyank (changes requested):

shahdivyank (approved):

PR #274, opened 2023-08-16
checkt recipient

### LINKED ISSUE #239: Add To Check
image

- Make sure that some group is always selected when clicking send email

menthy-wu:
image

### DIFF
--- a/src/components/Admin/Admissions.jsx
+++ b/src/components/Admin/Admissions.jsx
@@ -65,7 +65,7 @@ const Admissions = () => {
       return;
     }
     console.log(Object.entries(filters));
-    if (Object.entries(filters).filter((filter) => filter[1]).length == 0) {
+    if (Object.entries(filters).filter(([key, value]) => value).length == 0) {
       toast("❌ Please specify recipient!");
       return;
     }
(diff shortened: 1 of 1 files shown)


shahdivyank (changes requested):
this should boil down to 2 types of errors

1. no filter is chosen
2. a filter is chosen but there are no recipients under it cause then it will send an email to no one

so add another if statement to account for that. the current error message should say something like no emails to send to or something

src/components/Admin/Admissions.jsx
```diff
@@ -64,8 +64,15 @@ const Admissions = () => {
       toast("❌ Please add a body!");
       return;
     }
+    if (!filters.accepted && !filters.rejected) {
```
shahdivyank: map through the filters and check the size of the array. allows for us to put more filters without worrying bout code changes

Object.entries(values).filter(([key, value]) => value)

shahdivyank (changes requested):

src/components/Admin/Admissions.jsx
```diff
@@ -64,8 +64,15 @@ const Admissions = () => {
       toast("❌ Please add a body!");
       return;
     }
+    if (Object.keys(filters).filter((filter) => filters[filter]).length == 0) {
```
shahdivyank: use entries

shahdivyank: Object.entries

shahdivyank (approved):

PR #278, opened 2023-08-16
Create Feedback.jsx

### LINKED ISSUE #198: Create Hackathon Feedback Form
- Create a new form that asks for general feedback regarding the hackathon 
- Come up with your own questions
- Use the format specified by volunteers/mentors/register

sophiiiapapaya:
Not sure if the components are correctly imported/defined.

Need corrections for the name, type, value etc. for each question

### DIFF
--- a/src/components/Forms/Feedback.jsx
+++ b/src/components/Forms/Feedback.jsx
@@ -1,0 +1,124 @@
+"use client";
+
+import React, { useState } from "react";
+import Select from "@/components/Select";
+import { Majors, Grades } from "@/data/Register";
+import Input from "@/components/Input";
+import { Row, Col } from "react-bootstrap";
+import Button from "@/components/Forms/Button";
+import { Helper } from "@/data/User";
+
+const Mentor = () => {
+  const [mentor, setMentor] = useState(Helper);
+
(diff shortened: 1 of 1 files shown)


src/components/Forms/Feedback.jsx
```diff
@@ -0,0 +1,124 @@
...
+import React, { useState } from "react";
+import Select from "@/components/Select";
+import { Majors, Grades } from "@/data/Register";
+import Input from "@/components/Input";
+import { Row, Col } from "react-bootstrap";
+import Button from "@/components/Forms/Button";
+import { Helper } from "@/data/User";
+
+const Mentor = () => {
+  const [mentor, setMentor] = useState(Helper);
```
shahdivyank: don't call it mentor, make sure to change it for all feedback stuff. we also don't need to keep track of users since it can be an anonymous form

shahdivyank (changes requested):

src/components/Forms/Feedback.jsx
sophiiiapapaya: @shahdivyank I made the changes

shahdivyank (approved):

PR #289, opened 2023-08-19
Refactored all object.keys to use object.entries

### LINKED ISSUE #257: Refactor to use `Object.entries`
image

- Using object.entries allows us to access the value directly instead of having to use object.keys and then accessing the element via the provided key
- In ALL instances where we use object.keys and eventually need the value, instead simply use object.entries

seanquiambao:
Toolbar.jsx and Filter.jsx functions the same as before.


shahdivyank:
Not needed anymore

PR #319, opened 2023-08-21
form scroll

### LINKED ISSUE #301: BUG: Form Background
image

- Currently the background is only gray for 100vh, when it should be minimum of 100vh or the entire height of the container including the form

menthy-wu:
from
image
to
image

### DIFF
--- a/src/components/FormPage.jsx
+++ b/src/components/FormPage.jsx
@@ -1,0 +1,13 @@
+"use client";
+import React from "react";
+
+const FormPage = ({ title, children, restrictions }) => {
+  return (
+    <div className="w-full flex justify-center h-full overflow-scroll">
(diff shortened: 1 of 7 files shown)


menthy-wu:
Lol i did it and then realize that there's an issue for it

src/components/FormPage.jsx
```diff
@@ -0,0 +1,13 @@
+"use client";
+import React from "react";
+
+const FormPage = ({ title, children, restrictions }) => {
```
shahdivyank: - remove restrictions since its not being used
- call the component Form.jsx
- Remove the use client, this doesn't need it, its children need it

shahdivyank (approved):

PR #368, opened 2023-08-22
Extracted volunteer desccription and requirements

### LINKED ISSUE #249: Extract Volunteer Data
Within the page, there is tons of text that will change based on hackathons and their requirements, we want to put this information in a easily editable file. Things include 
image
image

- Both of these requirements should be put into a Volunteers.jsx file that lives in the Data folder. Make sure to import them back into their respective pages

seanquiambao:
image

extracted some of the replaceable keywords in the description and put them all into an object in Volunteer.js. for some reason, running format produces `{" "}` in the description.


shahdivyank:
extract the entire text, you don't have to extract just the variables, hackathons may want to change the delivery of the message and we should be able to account for that.

src/data/Volunteers.js
```diff
@@ -0,0 +1,14 @@
+export const Description = [
+  "Hello! Rose Hack is coming January 14-15, 2023. Thank you for your interest in Rose Hack, UC Riverside’s women-centric hackathon founded by the female leaders of SWE and WINC!Volunteers are essential to our hackathon in helping run the entire event. Responsibilities include helping set up, tech support, clean up, distributing swag, etc. Meals are also provided during breakfast, lunch, and dinner times. If you are interested in joining the Rose Hack Team as a volunteer, please fill out this quick interest form below! \n",
```
shahdivyank: don't need the "\n", if you need an extra line, then integrate it within the p tag

shahdivyank (approved):

PR #377, opened 2023-08-23
Menthy/upload

### LINKED ISSUE #342: Add File Type Enhancement to Upload
image
image

- Within the upload component, add an attribute to the input tag to only accept certain types of files which can be determined by the parent
- Register form will only take pdf
- judges form will only take png/jpeg/jpg

### LINKED ISSUE #343: Add File Size Enhancement to Upload Component
- Parents should be able to control the file size limit
- For register page, since we are working with PDFs, the limit should be 150KB
- For the judges photo, the limit should be 1MB 
- We can later refine these values
- Do not need to worry about sending the user any message

### LINKED ISSUE #344: User Error Message Enhancement
- When the user uploads a file, do a check for the file size and if it is too big, then reject it and send them an error
- The same applies for the file type

menthy-wu:
image

### DIFF
--- a/src/components/Forms/Judge.jsx
+++ b/src/components/Forms/Judge.jsx
@@ -151,8 +151,8 @@ const Judge = () => {
               user={judge}
               setUser={setJudge}
               text="Upload Photo"
-              maxSize={{ MB: 1 }}
-              typeLimit={["png", "jpg", "jpeg"]}
+              maxSize={[1, "MB"]}
+              types={["png", "jpg", "jpeg"]}
             />
           </Col>
           <Col xl={12}>
--- a/src/components/Forms/Judge.jsx
+++ b/src/components/Forms/Judge.jsx
@@ -152,7 +152,7 @@ const Judge = () => {
               setUser={setJudge}
               text="Upload Photo"
               sizeLimit="1 MB"
-              typeLimit=".png,.jpg"
+              typeLimit=".png,.jpg,.jpeg"
             />
           </Col>
           <Col xl={12}>
--- a/src/components/Forms/Judge.jsx
+++ b/src/components/Forms/Judge.jsx
@@ -152,7 +152,7 @@ const Judge = () => {
               setUser={setJudge}
               text="Upload Photo"
               maxSize="1 MB"
-              typeLimit=".png,.jpg,.jpeg"
+              typeLimit={["png", "jpg", "jpeg"]}
             />
           </Col>
           <Col xl={12}>
--- a/src/components/Forms/Judge.jsx
+++ b/src/components/Forms/Judge.jsx
@@ -151,7 +151,7 @@ const Judge = () => {
               user={judge}
               setUser={setJudge}
               text="Upload Photo"
-              maxSize="1 MB"
+              maxSize={{ MB: 1 }}
               typeLimit={["png", "jpg", "jpeg"]}
             />
           </Col>
--- a/src/components/Forms/Judge.jsx
+++ b/src/components/Forms/Judge.jsx
@@ -151,7 +151,7 @@ const Judge = () => {
               user={judge}
               setUser={setJudge}
               text="Upload Photo"
-              sizeLimit="1 MB"
+              maxSize="1 MB"
               typeLimit=".png,.jpg,.jpeg"
             />
           </Col>
--- a/src/components/Forms/Judge.jsx
+++ b/src/components/Forms/Judge.jsx
@@ -151,6 +151,8 @@ const Judge = () => {
               user={judge}
               setUser={setJudge}
               text="Upload Photo"
+              sizeLimit="1 MB"
+              typeLimit=".png,.jpg"
             />
           </Col>
           <Col xl={12}>
--- a/src/components/Forms/Register.jsx
+++ b/src/components/Forms/Register.jsx
@@ -154,6 +154,7 @@ const Register = () => {
               setUser={setUser}
               text="Upload Resume"
               sizeLimit="150 KB"
+              typeLimit=".pdf"
             />
           </Col>
           <Col xl={12}>
--- a/src/components/Forms/Register.jsx
+++ b/src/components/Forms/Register.jsx
@@ -153,8 +153,8 @@ const Register = () => {
               user={user}
               setUser={setUser}
               text="Upload Resume"
-              maxSize="150 KB"
-              typeLimit=".pdf"
+              maxSize={{ KB: 150 }}
+              typeLimit={["pdf"]}
             />
           </Col>
           <Col xl={12}>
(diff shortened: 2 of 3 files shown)


src/components/Forms/Judge.jsx
```diff
@@ -151,6 +151,8 @@ const Judge = () => {
               user={judge}
               setUser={setJudge}
               text="Upload Photo"
+              sizeLimit="1 MB"
+              typeLimit=".png,.jpg,.jpeg"
```
shahdivyank: call it just types and make it an array, its slightly more processing for us, but its easier to use

src/components/Forms/Judge.jsx
```diff
@@ -151,6 +151,8 @@ const Judge = () => {
               user={judge}
               setUser={setJudge}
               text="Upload Photo"
+              sizeLimit="1 MB"
```
shahdivyank: call it maxSize

src/components/Forms/Upload.jsx
```diff
@@ -1,14 +1,29 @@
 import React, { useState } from "react";
 import { BsUpload } from "react-icons/bs";
 import { FaFilePdf, FaTimes } from "react-icons/fa";
-
-const Upload = ({ field, user, setUser, text }) => {
+import { toast } from "react-hot-toast";
+const bytesArr = {
```
shahdivyank: call this just bytes, its not an arr, its an object

src/components/Forms/Register.jsx
```diff
@@ -153,6 +153,8 @@ const Register = () => {
               user={user}
               setUser={setUser}
               text="Upload Resume"
+              sizeLimit="150 KB"
```
shahdivyank: make this an easier format so there's less confusing code when splitting it. u could try something like:

```
maxSize = {
    {
        K: 150
    }
}
``

Comment (shahdivyank) [reply src/components/Forms/Register.jsx]: here you can use the key in the bytes obj and then the value as how much to multiply by

File: src/components/Forms/Upload.jsx [hunk 5]
```diff
@@ -1,14 +1,29 @@
...
+  GB: 1073741824,
+};
+const getSize = (sizeLimit) =>
+  sizeLimit.split(" ")[0] * bytesArr[sizeLimit.split(" ")[1]];
+const Upload = ({ field, user, setUser, text, sizeLimit, typeLimit }) => {
   const [file, setFile] = useState(null);
   const [uploading, setUploading] = useState(false);
 
   const handleInput = async (e) => {
+    const size = getSize(sizeLimit);
```
shahdivyank, line 18: what is the point of this?

src/components/Forms/Upload.jsx
```diff
@@ -1,14 +1,29 @@
...
+    if (!size) {
```
shahdivyank: what is the point of this ?

src/components/Forms/Upload.jsx
```diff
@@ -1,14 +1,29 @@
...
+      console.error("invalid upload sizeLimit input");
+      return;
+    }
     setUploading(true);
-    setFile(e.target.files[0]);
+    if (e.target.files[0].size > size) {
```
shahdivyank: what happens if the second file is too big, but the first file is okay?

shahdivyank:
It should handle multiple file upload as well

src/components/Forms/Judge.jsx
```diff
@@ -151,6 +151,8 @@ const Judge = () => {
               user={judge}
               setUser={setJudge}
               text="Upload Photo"
+              maxSize="1 MB"
```
shahdivyank: u missed a spot

src/components/Forms/Upload.jsx
```diff
@@ -1,14 +1,34 @@
...
+import { toast } from "react-hot-toast";
+const bytes = {
+  B: 1,
+  KB: 1024,
+  MB: 1048576,
+  GB: 1073741824,
+};
+const getSize = (maxSize) => {
+  let size = 0;
+  Object.entries(maxSize).forEach(([byte, num]) => (size += bytes[byte] * num));
```
shahdivyank: just return this directly, if size gets returned as 0 then we have issues in our codebase since this is internally handled

src/components/Forms/Upload.jsx
```diff
@@ -1,14 +1,34 @@
...
+  return size;
+};
+const getType = (typeLimit) => {
+  let typeStr = "";
+  typeLimit.forEach((type) => (typeStr += `.${type},`));
```
shahdivyank: same here, return the array directly

shahdivyank: can this be done using join?

src/components/Forms/Upload.jsx
```diff
@@ -1,14 +1,34 @@
...
+  return typeStr;
+};
+const Upload = ({ field, user, setUser, text, maxSize, typeLimit }) => {
```
shahdivyank: call it just types

src/components/Forms/Upload.jsx
```diff
@@ -1,14 +1,34 @@
...
   const [file, setFile] = useState(null);
   const [uploading, setUploading] = useState(false);
 
   const handleInput = async (e) => {
     setUploading(true);
-    setFile(e.target.files[0]);
+    if (e.target.files[0].size > getSize(maxSize)) {
+      let size = "";
```
shahdivyank: size shouldn't be "" so we can just use the size provided by the object entires

src/components/Forms/Upload.jsx
```diff
@@ -1,14 +1,34 @@
...
+  KB: 1024,
+  MB: 1048576,
+  GB: 1073741824,
+};
+const getSize = (maxSize) => {
+  let size = 0;
+  Object.entries(maxSize).forEach(([byte, num]) => (size += bytes[byte] * num));
+  return size;
+};
+const getType = (typeLimit) => {
```
shahdivyank: make this just a variable, if its not going to be called more than 1 time

src/components/Forms/Upload.jsx
```diff
@@ -1,14 +1,24 @@
...
+import { toast } from "react-hot-toast";
+const bytes = {
+  B: 1,
+  KB: 1024,
+  MB: 1048576,
+  GB: 1073741824,
+};
+const getSize = (maxSize) => bytes[maxSize[1]] * maxSize[0];
+const getType = (types) => "." + types.join(",.");
+const Upload = ({ field, user, setUser, text, maxSize, types }) => {
```
shahdivyank, line 14: add a newline
```suggestion

const Upload = ({ field, user, setUser, text, maxSize, types }) => {
```

shahdivyank (approved):

PR #402, opened 2023-08-24
Fixed Default Edit

### LINKED ISSUE #398: Input Default Editable
image
- The input component for the forms and user dashboard are the same
- You should give Input another prop to indicate if it is editable by default
- make sure both register form and user dashboard works

rfairooz:
Dashboard works like before
Screenshot 2023-08-25 at 1 58 35 PM
Screenshot 2023-08-25 at 1 58 15 PM
All Forms are now editable by default.

Screenshot 2023-08-25 at 2 01 52 PM

Addressed Comments

### DIFF
--- a/src/components/Forms/Register.jsx
+++ b/src/components/Forms/Register.jsx
@@ -55,7 +55,6 @@ const Register = () => {
               maxLength={50}
               user={user}
               setUser={setUser}
-              editable={true}
             />
           </Col>
           <Col xl={6}>
@@ -68,7 +67,6 @@ const Register = () => {
               maxLength={50}
               user={user}
               setUser={setUser}
-              editable={true}
             />
           </Col>
           <Col xl={6}>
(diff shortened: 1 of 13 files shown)


shahdivyank:
can you make the default in the true in the params and then if its false then specify. It seems to be true in most cases so in the component header of the input u can do
```
const Input = ({ something, something, editable = true }) => {
``

File: src/components/Forms/Register.jsx [hunk 1]
```diff
@@ -55,6 +55,7 @@ const Register = () => {
               maxLength={50}
               user={user}
               setUser={setUser}
+              editable={true}
```
shahdivyank: you can remove these lines now that the default is true. make sure to say its false in case u need that

rfairooz: ah shoot sorry missed that one

rfairooz: ok fixed register. All other forms alr fixed in prev commit

shahdivyank (approved):

PR #406, opened 2023-08-26
QR code Timestamp

### LINKED ISSUE #266: Append Timestamp to QR code
image

- The QR code should refresh every 5 seconds 
- This would require putting the timestamp in the QR code itself 
- We can append it to the end with a special character (we can also integrate throughout the sequence so its harder to read)

rfairooz:
Removed timezone calculations. My time is 1:32 PM, shown as 20:32. Lmk if that is what you required.  
Screenshot 2023-08-28 at 1 34 58 PM


src/components/User/CheckIn.jsx
```diff
@@ -36,7 +36,12 @@ const CheckinPage = () => {
...
         </div>
         <div className="bg-white w-2/3 h-1/3 lg:h-5/6 flex justify-center items-center flex-col rounded-lg m-auto">
-          <QRCodeSVG value={session.user.id} className="w-2/3 h-2/3" />
+          <QRCodeSVG
+            value={`${session.user.id}_${date.toLocaleTimeString("en-US", {
+              hour12: true,
+            })}`}
+            className="w-2/3 h-2/3"
+          />
           {date.toLocaleTimeString("en-US", { hour12: true })}
```
shahdivyank: remove the live time stamp here

src/components/User/CheckIn.jsx
```diff
@@ -36,7 +36,12 @@ const CheckinPage = () => {
           <p className="text-base">{session.user.email}</p>
         </div>
         <div className="bg-white w-2/3 h-1/3 lg:h-5/6 flex justify-center items-center flex-col rounded-lg m-auto">
-          <QRCodeSVG value={session.user.id} className="w-2/3 h-2/3" />
+          <QRCodeSVG
+            value={`${session.user.id}_${date.toLocaleTimeString("en-US", {
```
shahdivyank: make the separating character &

shahdivyank: make sure the timestamp doesn't have any spaces and store it in iso format

src/components/User/CheckIn.jsx
```diff
@@ -36,8 +36,12 @@ const CheckinPage = () => {
           <p className="text-base">{session.user.email}</p>
         </div>
         <div className="bg-white w-2/3 h-1/3 lg:h-5/6 flex justify-center items-center flex-col rounded-lg m-auto">
-          <QRCodeSVG value={session.user.id} className="w-2/3 h-2/3" />
-          {date.toLocaleTimeString("en-US", { hour12: true })}
+          <QRCodeSVG
+            value={`${session.user.id}&${date.toLocaleTimeString("en-US", {
```
shahdivyank: toISOString

rfairooz: gotcha. you wanted both date & time right? like the pic above?

src/components/User/CheckIn.jsx
```diff
@@ -36,8 +36,12 @@ const CheckinPage = () => {
           <p className="text-base">{session.user.email}</p>
         </div>
         <div className="bg-white w-2/3 h-1/3 lg:h-5/6 flex justify-center items-center flex-col rounded-lg m-auto">
-          <QRCodeSVG value={session.user.id} className="w-2/3 h-2/3" />
-          {date.toLocaleTimeString("en-US", { hour12: true })}
+          <QRCodeSVG
+            value={`${session.user.id}&${date.toISOString("en-US", {
```
shahdivyank: dont need the custom formatting stuff

rfairooz: removed formatting but the iso time shows wrong time since it doesn't consider the time zone. I had to calculate offset and then convert to iso. Lmk if that looks okay.

src/components/User/CheckIn.jsx
```diff
@@ -13,11 +13,15 @@ const CheckinPage = () => {
   useEffect(() => {
     const timer = setInterval(() => {
       setDate(new Date());
-    }, 1000);
+    }, 5000);
 
     return () => clearInterval(timer);
   }, []);
 
+  const isoTime = new Date(
```
shahdivyank: make this inline

shahdivyank: what is the subtraction and multiplication for? If its for timezone we can remove it, we can base it on meridian time

shahdivyank (approved):

src/components/User/CheckIn.jsx
shahdivyank: Default is meridian time, which is UTC timezone so its fine

menthy-wu (approved):

PR #408, opened 2023-08-27
Changed accept and reject statuses to past tense

### LINKED ISSUE #379: Participant Status Past Tense
image

- Status should be accepted or rejected, having it present tense implies some actions can be done if clicked

seanquiambao:
pasttense

this change should not affect all of the toolbar functions, filter functions, etc... should only changed the text of tags on the table and the filters.


src/components/Admin/Filters.jsx
```diff
@@ -36,7 +36,8 @@ const Filters = ({ filters, setFilters, setObjects, objects, input }) => {
             } cursor-pointer flex items-center w-fit m-0`}
           >
             <p className="my-0 mx-1 px-2 py-[2px] whitespace-nowrap">
-              {filter}
+              {filter +
+                (filter === "accept" || filter === "reject" ? "ed" : "")}
```
shahdivyank: we will set up a standard across for all pages, so we dont have to do any conditional stuff to account for different pages

shahdivyank (approved):

rfairooz (approved):

PR #425, opened 2023-08-29
create a register.jsx under the data folder to export register data

### LINKED ISSUE #251: Extract Register Data
Within the register form, there is static data that will change between hackathons and we want an easy interface to work with them 
image

- Create a Register.jsx file under the data folder which holds such static information

twang0323:
image


src/data/Register.jsx
```diff
@@ -0,0 +1,9 @@
+export const requirphotography = {
```
shahdivyank: create a single requirement array and have an array of objects, reference https://github.com/acm-ucr/hackathon-website/blob/dev/src/data/Volunteers.js#L5

twang0323: image
like this?

src/data/Register.jsx
```diff
@@ -0,0 +1,10 @@
+export const Registration = {
```
shahdivyank: make sure to import it and use it as well

twang0323: image
I imported it into form/register and used it in the second line

shahdivyank: make sure to push these changes as well, doesn't look like they are pushed

shahdivyank (approved):

rfairooz (approved):

PR #431, opened 2023-08-30
fix header

menthy-wu:


src/components/Admin/Table.jsx
```diff
@@ -60,7 +60,7 @@ const Table = ({ headers, setHeaders, empty, setObjects, objects }) => {
             key={index}
             className="font-semibold text-white flex items-center p-0"
           >
-            {header.text}
+            {header.text ? header.text : ""}
```
shahdivyank: what happens if we just leave it null?

shahdivyank (approved):

rfairooz (approved):

PR #437, opened 2023-09-01
Disable drop drop menu when not in edit mode

### LINKED ISSUE #262: Disable Drop Down Menu
image

- When the edit option is not enabled, then the dropdown arrow should not show up
- This may require disabling the react bootstrap icon and using our own instead (totally ok!)

SafeDuck:
image

### DIFF
--- a/src/components/Select.jsx
+++ b/src/components/Select.jsx
@@ -13,14 +13,25 @@ const Select = ({
     <div className="mt-3">
       <p className="mb-1">{title}</p>
       <Dropdown className="w-full m-0">
-        <Dropdown.Toggle
-          id="dropdown-toggle"
-          className={`!bg-white ${
-            user[field] ? "text-black" : "!text-hackathon-placeholder"
-          } w-full !text-left !border-x-0 !border-t-0 !border-b-2 !rounded-none !border-black`}
-        >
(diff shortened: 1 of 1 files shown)


src/components/Select.jsx
```diff
@@ -13,14 +13,25 @@ const Select = ({
...
       <Dropdown className="w-full m-0">
-        <Dropdown.Toggle
-          id="dropdown-toggle"
-          className={`!bg-white ${
-            user[field] ? "text-black" : "!text-hackathon-placeholder"
-          } w-full !text-left !border-x-0 !border-t-0 !border-b-2 !rounded-none !border-black`}
-        >
-          {user[field] || placeholder}
-        </Dropdown.Toggle>
+        {editable ? (
```
shahdivyank, line 16: see if there is a way to keep the dropdown toggle but just remove the icon and its functionality conditionally

SafeDuck: but when you click the pencil, the dropdown toggle comes back and conditionally operates only when in edit mode

SafeDuck: image

shahdivyank (approved):

menthy-wu (approved):

PR #565, opened 2023-09-07
Created test cases for volunteer searching

### LINKED ISSUE #521: Volunteers Searching E2E Test
- When searching for a user, the appropriate results should show up

seanquiambao:
image


shahdivyank:
Add a test where multiple results come up due to a partially typed input or multiple names being the same

seanquiambao:
> Add a test where multiple results come up due to a partially typed input or multiple names being the same

I added the test case, lmk if there's something I need to fix or change!

cypress/e2e/admin/volunteers/Search.cy.js
```diff
@@ -0,0 +1,35 @@
...
+    cy.get(`[data-cy="${volunteerList[0].uid}"]`).should("exist");
+  });
+
+  it("Search Partial Results", () => {
+    cy.get('[data-cy="toolbar"]')
+      .find("input")
+      .type(multiple[0].name.split(" ")[0]);
+    cy.get('[data-cy="toolbar"]').find("form").submit();
+
+    multiple.forEach((particpant) =>
```
shahdivyank: - spelling
- what r the names of the first 3 participants? since we r checking for first name on them

seanquiambao: >     * what r the names of the first 3 participants? since we r checking for first name on them
- `multiple` stores all of John Cenas in the volunteer data. So the test will search for "John" in the search bar.

PR #589, opened 2023-09-10
Add volunteers filtering test

### LINKED ISSUE #504: Volunteers Filtering E2E Test
- Make tests for the filtering
- When a filter is deselected then the updated objects should replicate accordingly

suhanic44:
Screen Shot 2023-09-09 at 6 26 31 PM


cypress/e2e/admin/volunteers/Filter.cy.js
```diff
@@ -0,0 +1,112 @@
+import { volunteerList } from "../../../../src/data/mock/Volunteers";
```
SafeDuck: mock folder no longer exists, might need to change this line to 
`import volunteerList from "../../../fixtures/Volunteers.json";`
or might need to pull from dev.
Divyank will confirm.

PR #605, opened 2023-09-11
toast when qr is invalid, display no data

### LINKED ISSUE #452: Only Check In on Valid Data
image

- If there is no valid QR code present, then give a react toast indicating no valid QR code and do not display any data

SafeDuck:
image
image

### DIFF
--- a/src/components/Admin/Services/CheckInfo.jsx
+++ b/src/components/Admin/Services/CheckInfo.jsx
@@ -4,7 +4,7 @@ import toast from "react-hot-toast";
 
 const CheckInfo = ({ info, events }) => {
   if (!info) {
-    toast("❌ Not a Valid QR Code!");
+    toast("❌ Invalid QR Code!");
   }
   return info ? (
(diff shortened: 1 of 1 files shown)


src/components/Admin/Services/CheckInfo.jsx
```diff
@@ -1,7 +1,11 @@
 import Image from "next/image";
 import { AiOutlineCheck } from "react-icons/ai";
+import toast from "react-hot-toast";
 
 const CheckInfo = ({ info, events }) => {
+  if (!info) {
+    toast("❌ Not a Valid QR Code!");
```
shahdivyank: Invalid

shahdivyank:
```suggestion
    toast("❌ Invalid QR Code!");
```

shahdivyank (approved):

rfairooz (approved):

PR #619, opened 2023-09-14
pdf viewer fixes, use iframe instead of img, remove pdf toolbar

SafeDuck:
image


SafeDuck:
too many changes made, need to pull from dev

PR #644, opened 2023-09-17
check empty

### LINKED ISSUE #629: Fix the Feedback Questions
- extract the questions into the data folder
- You do not need the helper variable in order to implement the feedback form
- Make sure that the responses update the useState properly
- Upon submission it should console log the responses

### LINKED ISSUE #630: Add Feedback Form Validation
- We will make the form a ALL OR NOTHING form, so they can't fill the form partially since it would mean that they could send an empty form wasting storage
- Make sure every question is filled out (non empty)

menthy-wu:
image

### DIFF
--- a/src/components/forms/Feedback.jsx
+++ b/src/components/forms/Feedback.jsx
@@ -5,12 +5,17 @@ import Input from "@/components/Input";
 import Col from "react-bootstrap/Col";
 import Row from "react-bootstrap/Row";
 import Button from "@/components/Button";
-import { HELPER } from "@/data/forms/Helper";
+import { feedbackInfo } from "@/data/forms/Helper";
+import toast from "react-hot-toast";
 
 const Feedback = () => {
-  const [feedback, setFeedback] = useState(HELPER);
+  const [feedback, setFeedback] = useState(feedbackInfo);
 
   const handleSubmit = () => {
+    if (Object.values(feedback).some((value) => value === "")) {
+      toast("❌ Please complete all fields!");
+      return;
+    }
     console.log(feedback);
   };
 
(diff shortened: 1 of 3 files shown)


src/data/forms/Helper.js
```diff
@@ -59,3 +59,11 @@ export const AVAILABILITY = {
     text: "Sunday Night",
   },
 };
+export const feedbackInfo = {
+  question1: "",
```
shahdivyank: come up with the questions pls

src/data/forms/Helper.js
```diff
@@ -59,3 +59,11 @@ export const AVAILABILITY = {
     text: "Sunday Night",
   },
 };
+export const feedbackInfo = {
+  helpful: "",
```
shahdivyank: write the questions out 😭

src/data/forms/Helper.js
```diff
@@ -59,3 +59,11 @@ export const AVAILABILITY = {
     text: "Sunday Night",
   },
 };
+export const feedbackInfo = {
```
shahdivyank: call it FEEDBACK and make sure its all caps, also this should be in its own file, helper file is mostly for mentor and volunteer common attributes

src/components/forms/Feedback.jsx
```diff
@@ -64,7 +69,7 @@ const Feedback = () => {
           </Col>
           <Col xl={12}>
             <Input
```
shahdivyank, line 71: rating should be a radio

src/components/forms/Feedback.jsx
```diff
@@ -76,7 +81,7 @@ const Feedback = () => {
...
-              name="question5"
+              name="How did you find the event?"
```
shahdivyank, line 84: This should be a select

PR #657, opened 2023-09-18
Menthy/form

menthy-wu:


src/app/forms/Form.jsx
```diff
@@ -0,0 +1,124 @@
...
+                  </>
+                ) : field.input === "radio" ? (
+                  <Radio
+                    text={field.text}
+                    options={field.options}
+                    field={field.field}
+                    user={object}
+                    setUser={setObject}
+                  />
+                ) : field.input === "textarea" ? (
```
shahdivyank: do the && since only one of these options is possible, we dont need the chaining

src/components/forms/Admin.jsx
```diff
@@ -1,127 +1,26 @@
...
-  const handleRequirementsCheckbox = (key, value) => {
-    setRequirements({
-      ...requirements,
-      [key]: { state: !value.state, text: value.text },
-    });
-  };
-
+  const { data: session } = useSession();
+  const [admin, setAdmin] = useState({
+    name: session.user.name,
```
shahdivyank, line 11: this info should be collected in the end, otherwise we store it and do nothing until the end, check volunteer

PR #672, opened 2023-09-20
Select component test

### LINKED ISSUE #495: Select Component Test
/components/Select

SafeDuck:
image


cypress/component/Select.cy.jsx
```diff
@@ -0,0 +1,55 @@
+import Select from "@/components/Select";
+describe("Select", () => {
+  const options = ["Option 1", "Option 2", "Option 3"];
+  const field = "field";
+  const placeholder = "Select an option";
+  const title = "Select an option";
+
+  it("Renders", () => {
+    const setUserStub = cy.stub();
```
shahdivyank: make this a hook that u can use

cypress/component/Select.cy.jsx
```diff
@@ -0,0 +1,55 @@
...
+
+    cy.get('[data-cy="toggle"]').should("exist");
+    cy.get('[data-cy="menu"]').should("not.exist");
+  });
+
+  it("Show dropdown menu", () => {
+    const setUserStub = cy.stub();
+
+    const Parent = () => {
+      const user = { [field]: "" };
```
shahdivyank: use react hooks and u will be able to test the actual data manipulation

cypress/component/Select.cy.jsx
```diff
@@ -0,0 +1,53 @@
...
+
+describe("Select", () => {
+  const options = ["Option A", "Option B", "Option C"];
+  const field = "field";
+  const placeholder = "Select an option";
+  const title = "Title";
+
+  it("Renders", () => {
+    const Parent = () => {
+      const [user, setUser] = useState({ [field]: "" });
```
shahdivyank: dont make the field a variable, use a string, also use a better attribute, something more relevant like "name"

cypress/component/Select.cy.jsx
```diff
@@ -0,0 +1,52 @@
...
+      );
+    };
+
+    cy.mount(<Parent />);
+
+    cy.get('[data-cy="toggle"]').should("exist");
+    cy.get('[data-cy="menu"]').should("not.exist");
+  });
+
+  it("Show dropdown menu", () => {
```
shahdivyank:
```suggestion
  it("Show Dropdown", () => {
```

cypress/component/Select.cy.jsx
```diff
@@ -0,0 +1,52 @@
+import Select from "@/components/Select";
```
shahdivyank, line 1: add a test that selects an option and then make sure that update actually happened

SafeDuck: isn't that already in the e2e tests for sort?

shahdivyank: comp test should make sure that the entire workflow for a comp works. testing the select values and stuff is comp test cause it is handled internally by the comp and not by a parent comp

src/components/Select.jsx
```diff
@@ -0,0 +1,58 @@
+import Dropdown from "react-bootstrap/Dropdown";
+
+const Select = ({
```
shahdivyank: uh there is a whole component being added

SafeDuck: ah sheet ma bad, weird merging. Menthy suggested to delete it from my branch

shahdivyank (approved):

PR #678, opened 2023-09-21
Menthy/mentortests

menthy-wu:


src/components/admin/dashboards/SortIcon.jsx
```diff
@@ -34,15 +35,26 @@ const SortIcon = ({
...
-      objects.sort((a, b) => {
-        if (state === "up") {
-          return a[name] > b[name] ? -1 : 1;
-        } else if (state === "down") {
-          return b[name] > a[name] ? -1 : 1;
-        }
-      })
-    );
+    if (name === "status")
+      setObjects(
```
shahdivyank: combine into one set objects

src/components/admin/dashboards/SortIcon.jsx
```diff
@@ -37,9 +38,11 @@ const SortIcon = ({
     setObjects(
       objects.sort((a, b) => {
         if (state === "up") {
-          return a[name] > b[name] ? -1 : 1;
+          if (name === "status") return a[name][page] > b[name][page] ? -1 : 1;
+          else return a[name] > b[name] ? -1 : 1;
```
shahdivyank: u dont need the else, u can just return regardless

PR #684, opened 2023-09-21
Menthy/dropdown

menthy-wu:
image

### DIFF
--- a/src/components/admin/services/DropDown.jsx
+++ b/src/components/admin/services/DropDown.jsx
@@ -6,7 +6,7 @@ const Toggle = ({ option, onClick, show }) => {
   return (
     <button
       onClick={onClick}
-      className="my-1 bg-hackathon-gray-100 rounded-md relative z-50 w-full justify-between flex items-center px-3 py-1"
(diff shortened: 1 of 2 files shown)


src/components/admin/services/Toolbar.jsx
```diff
@@ -127,7 +127,7 @@ const Toolbar = ({ objects, setObjects, teams, setTeams }) => {
   return (
     <>
       <div className="flex items-center justify-between w-full"></div>
```
shahdivyank: this should be self closing, why is this even here?

src/components/admin/services/DropDown.jsx
```diff
@@ -6,7 +6,7 @@ const Toggle = ({ option, onClick, show }) => {
   return (
     <button
       onClick={onClick}
-      className="relative z-50 bg-white text-black w-full border-2 rounded-lg border-black justify-between flex items-center px-3 py-1"
+      className="my-1 bg-hackathon-gray-100 rounded-md relative z-50 w-full justify-between flex items-center px-3 py-1"
```
shahdivyank: why do we need 50? use 10, do we even need it at all?

shahdivyank: what's the relative for?

PR #688, opened 2023-09-23
struggling to make the the search bar :////

isabellecelo:
i was able to add the search bar to the school drop down, but it does not filter the schools yet


src/components/admin/services/DropDown.jsx
```diff
@@ -51,7 +51,7 @@ const Menu = ({ setOption, className, setOptions, options }) => {
             </Dropdown.Item>
           ))
       ) : (
-        <p className="px-3 text-gray-400">no events</p>
+        <p className="px-3 text-gray-400"></p>
```
menthy-wu: this is the text displace when there's no options, so it should be a customized text pass in dropdown as a prop

src/components/admin/services/DropDown.jsx
```diff
@@ -34,7 +34,7 @@ const Menu = ({ setOption, className, setOptions, options }) => {
       <input
         autoFocus
         className="mx-1.5 my-1 w-11/12 ring-0 outline-none px-2 py-1"
-        placeholder="Type to filter..."
+        placeholder="Search"
```
menthy-wu: this should be "Type to filter..."

menthy-wu:
For select component, it is too hard to explain in pr review, you can come to me and we can figure out how to fix it. discord: _wmx

src/components/Select.jsx
```diff
@@ -1,27 +1,51 @@
...
+      option ? "text-black" : "!text-hackathon-gray-200"
+    } w-full !text-left !border-x-0 !border-t-0 !border-b-2 !rounded-none !border-black`}
+  >
+    {option || placeholder}
+  </button>
+);
 
 const Select = ({
-  options,
+  defaultOptions,
```
shahdivyank: rename to default

src/components/Select.jsx
```diff
@@ -35,16 +59,27 @@ const Select = ({
...
-              >
-                {option}
-              </Dropdown.Item>
-            ))}
+          <Dropdown.Menu className="w-full !bg-hackathon-green-100 !border-none !rounded-none !p-0 overflow-y-auto max-h-[35vh]">
+            {searchable && (
+              <input
+                autoFocus
+                className="mx-1.5 my-1 w-11/12 ring-0 outline-none px-2 py-1 bg-hackathon-green-100"
+                placeholder="Search"
```
shahdivyank: lowercase s on search

src/components/Select.jsx
```diff
@@ -35,16 +59,27 @@ const Select = ({
...
+                onChange={handleInput}
+              />
+            )}
+            {options.map(
+              (option, index) =>
+                !option.hidden && (
```
shahdivyank: if ur only showing visible elements, then u can filter and then map, it will be faster

src/components/dynamic/Select.jsx
```diff
@@ -64,22 +64,21 @@ const Select = ({
...
-                    className=" hover:!bg-hackathon-green-200 !bg-hackathon-green-100 overflow-hidden"
-                    key={index}
-                    onClick={() => setUser({ ...user, [field]: option.name })}
-                  >
-                    {option.name}
-                  </Dropdown.Item>
-                )
-            )}
+            {options
+              .filter((option, index) => !option.hidden)
```
shahdivyank: u dont need the index here, since u will use the map one anyways

src/components/dynamic/Select.jsx
```diff
@@ -1,27 +1,51 @@
 import Dropdown from "react-bootstrap/Dropdown";
+import { useState } from "react";
+
+const Toggle = ({ option, onClick, placeholder }) => (
+  <button
+    onClick={onClick}
+    data-cy="select-toggle"
+    className={`!bg-white ${
+      option ? "text-black" : "!text-hackathon-gray-200"
+    } w-full !text-left !border-x-0 !border-t-0 !border-b-2 !rounded-none !border-black`}
```
shahdivyank: this is a button comp, u dont need ! for any of these

src/components/dynamic/Select.jsx
```diff
@@ -1,27 +1,51 @@
...
+  >
+    {option || placeholder}
+  </button>
+);
 
 const Select = ({
-  options,
+  defaultOptions,
```
shahdivyank: call this just options, that isn't being used anywhere

shahdivyank (approved):

PR #702, opened 2023-09-26
added status row to Messenger page

### LINKED ISSUE #682: Need 2 Filters
image

- Should be 2 lines, where top line is the group of users (participants, volunteers, mentors, admins and so on)
- The second line should be their status, since each user can have 3 diff status

fardinzam:
Screenshot 2023-09-25 at 7 56 48 PM

### DIFF
--- a/src/components/admin/services/Messenger.jsx
+++ b/src/components/admin/services/Messenger.jsx
@@ -9,6 +9,7 @@ import Button from "../Button";
 import toast from "react-hot-toast";
 import Upload from "./Upload";
 import { FILTERS } from "@/data/admin/Messenger";
+import { FILTERS_ } from "@/data/admin/Messenger2";
 
 const Messenger = () => {
   const [email, setEmail] = useState({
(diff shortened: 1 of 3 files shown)


src/data/admin/Messenger2.js
```diff
@@ -0,0 +1,5 @@
+export const FILTERS_ = {
```
shahdivyank, line 1: call it STATUSES and put in the same file as messenger

src/components/admin/services/Messenger.jsx
```diff
@@ -44,6 +48,16 @@ const Messenger = () => {
             input=""
           />
         </div>
+        <div className="flex items-center my-1">
+          <p className="text-lg font-extrabold mr-5 my-0">status:</p>
+          <Filters
+            filters={filters_}
+            setFilters={setFilters_}
+            setObjects={() => {}}
```
shahdivyank, line 56: need to be valid values

src/components/admin/services/Messenger.jsx
```diff
@@ -18,6 +19,8 @@ const Messenger = () => {
     files: [],
   });
   const [filters, setFilters] = useState(FILTERS);
+  const [filters_, setFilters_] = useState(FILTERS_);
```
shahdivyank, line 22: call it statuses

PR #784, opened 2023-10-11
restricts winner to qualify action and adds a popup

### LINKED ISSUE #718: Disable Winner -> Qualify Action
- This is a special case that could occur and would throw off the prizes page
- Currently winner status should only be controlled by prizes page
- If the user tries to change a teams status from winner to qualify manually, trigger a popup that says they cannot do that and send them to the prizes tab instead

twang0323:


twang0323:
image

src/components/dynamic/admin/dashboards/Toolbar.jsx
```diff
@@ -28,6 +28,19 @@ const Toolbar = ({
   const [toggle, setToggle] = useState(false);
 
   const onClick = (value) => {
+    const selectedWinners = objects.filter(
+      (obj) => obj.selected && obj.status === "winner"
+    );
+    if (value === "qualify" && selectedWinners.length > 0) {
+      setPopup({
+        title: "Status Change Restricted",
+        text: "Changing status from 'winner' to 'qualify' is restricted. You can check the Prize page for more information.",
```
shahdivyank: what happens when its not qualify, so they went from winner->pending. Also check the some function, it can return a boolean and do exactly what u need here. u rlly dont need a value check cause the only values u can change to will be non winner so it should raise a flag solely based on selectedWinners array

src/components/dynamic/admin/dashboards/Toolbar.jsx
```diff
@@ -28,6 +28,19 @@ const Toolbar = ({
   const [toggle, setToggle] = useState(false);
 
   const onClick = (value) => {
+    const selectedWinners = objects.some(
+      (obj) => obj.selected && obj.status === "winner"
+    );
+    if (selectedWinners) {
+      setPopup({
+        title: "Status Change Restricted",
+        text: "Changing status from is restricted. You can check the Prize page for more information.",
```
shahdivyank: from winner

shahdivyank: Changing status from winner is restricted. Please remove the prize for this team before changing their status

shahdivyank: there should be a link option that u need to enable? so that the button takes u to the prizes page

twang0323:
second fix

src/components/dynamic/admin/dashboards/Toolbar.jsx
```diff
@@ -134,23 +159,26 @@ const Toolbar = ({
...
-              ...popup,
-              visible: true,
-            })
-          }
+          onClick={() => setDeletePopup({ ...deletePopup, visible: true })}
           size={22.5}
           className="ml-5 text-hackathon-gray-300 hover:opacity-70 duration-150 hover:cursor-pointer"
         />
-        {popup.visible && (
+        {deletePopup.visible && (
```
shahdivyank: there should only be one popup, we should just be changing the content of the popup

twang0323: image
fixed code

src/components/dynamic/admin/Popup.jsx
```diff
@@ -1,7 +1,7 @@
 import Button from "./Button";
 import { LiaTimesSolid } from "react-icons/lia";
 
-const Popup = ({ popup, onClick, setPopup, text }) => {
+const Popup = ({ popup, onClick, setPopup, text, onCancel }) => {
```
shahdivyank: remove on cancel

shahdivyank: remove on cancel

src/components/dynamic/admin/services/judging/Toolbar.jsx
```diff
@@ -184,7 +184,7 @@ const Toolbar = ({ data, setData, judges, setJudges }) => {
         <Popup
           setPopup={setPopup}
           popup={popup}
-          onClick={() => router.push("/admin/judges")}
+          onClick={() => router.push("/admins/judges")}
```
shahdivyank: should be admin

shahdivyank (approved):

PR #812, opened 2023-10-17
fix alignment

JohnZ9865:
Fixed alignment --- I had to delete the last branch with the same name because of an git bug (I'll probably ask about it after meeting or something).
But yeah here's the proof: 
image
image

Edit: better pictures
image
image


shahdivyank:
attach issue

src/components/dynamic/admin/Table.jsx
```diff
@@ -94,7 +94,7 @@ const Table = ({
                   object.selected ? "bg-green-100" : "bg-white"
                 }`}
               >
-                <div className="flex justify-between items-center">
+                <div className="flex justify-start items-center">
```
shahdivyank, line 97: this change should not be necessary

JohnZ9865: But without it, it looks like this:
image

I initially thought so too and was confused, but Menthy suggested I add justify-start to accordion and then that worked.

shahdivyank (approved):

PR #820, opened 2023-10-21
added calendar shortcuts

### LINKED ISSUE #802: Add Calendar Keyboard Shortcuts
- When the user presses 'w' or 'm' it should switch to week and month view respectively

seanquiambao:
tried using a ternary operator, but i couldn't figure out how to handle default case since pressing any other key will change to the initial value of view.

used keycode to take into account of capital letters.

### DIFF
--- a/src/components/dynamic/admin/services/calendar/Events.jsx
+++ b/src/components/dynamic/admin/services/calendar/Events.jsx
@@ -18,11 +18,11 @@ const CalendarEvents = () => {
   const [view, setView] = useState("month");
 
   const handleShortcuts = (e) => {
-    switch (e.keyCode) {
-      case 77:
+    switch (e.key) {
+      case "m":
         setView("month");
         break;
-      case 87:
+      case "w":
         setView("week");
         break;
       default:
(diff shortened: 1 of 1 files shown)


src/components/dynamic/admin/services/calendar/Events.jsx
```diff
@@ -17,6 +17,19 @@ const CalendarEvents = () => {
   const [date, setDate] = useState(new Date());
   const [view, setView] = useState("month");
 
+  const handleShortcuts = (e) => {
+    switch (e.keyCode) {
```
shahdivyank: there is there a way to match it based on character like if e.character === "w" so we dont need to know the ASCII codes

seanquiambao: so would i use `e.code === "KeyW"` or `e.key === "w" || e.key === "W"`

seanquiambao: `e.code === "KeyW"` or just `e.key === "w" || e.key === "W"`?

src/components/dynamic/admin/services/calendar/Events.jsx
```diff
@@ -17,6 +17,19 @@ const CalendarEvents = () => {
...
+  const handleShortcuts = (e) => {
+    switch (e.key) {
+      case "m":
+        setView("month");
+        break;
+      case "w":
+        setView("week");
+        break;
+      default:
+        break;
```
shahdivyank:
```suggestion
```

shahdivyank (approved):

PR #840, opened 2023-10-23
added component testing??

### LINKED ISSUE #821: Enable Component Tests
- currently only e2e tests are being run, we need component tests enabled in Github Actions
- Fix up any broken component tests if necessary

seanquiambao:
removed `npm run start` in component step because it'll throw an EADDRINUSE error. `npm run build` was deleted  also to make make the action run faster.


cypress/fixtures/events.json
```diff
@@ -0,0 +1,13 @@
+{
+  "events": [
+    {
+      "name": "React Workshop"
```
shahdivyank, line 5: there should be an id on each of these, u can make random ids

.github/workflows/pull-request.yaml
```diff
@@ -60,14 +60,23 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
 
-      - name: Run Cypress
+      - name: Run Cypress - E2E Tests
```
shahdivyank: for now u can keep the same job and run both component and e2e in the same job

shahdivyank (approved):

PR #843, opened 2023-10-25
full width 832, made whole thing bigger and all in one line

JohnZ9865:
image

Changes made:
- for some reason the thing was a grid with 3 cols. I made it 6
- Also on page.js for messenger it was 11/12 where I highlighted I made it 12/12 
image


src/app/admin/messenger/page.js
```diff
@@ -10,7 +10,7 @@ const MessengerPage = () => {
         committees: 1,
       }}
     >
-      <div className="w-11/12 justify-center bg-hackathon-page z-0 h-screen">
+      <div className="w-12/12 justify-center bg-hackathon-page z-0 h-screen">
```
shahdivyank: w-full

src/components/dynamic/admin/services/Messenger.jsx
```diff
@@ -72,8 +72,10 @@ const Messenger = () => {
       <div className="flex flex-col pb-3 pt-4 h-full items-stretch justify-between">
         <Title title="Messenger" />
         <div className="flex items-center my-1">
-          <p className="text-lg font-extrabold mr-5 my-0">to:</p>
-          <Filters filters={filters} setFilters={setFilters} />
+          <p className="text-lg font-extrabold mr-1 my-0">to:</p>
+          <div className="w-full ">
```
shahdivyank: Is the div needed here?

JohnZ9865:
Actually I think you can ignore this one. because the other issue, 831, already fixes everything here I believe. I think the two 2 identical.

PR #849, opened 2023-10-26
moved 404 and 500 errors into error component

### LINKED ISSUE #834: Aggregate Errors
- Currently there are 3 different Error Implementations: error.js, not-found.js, Error Component 
- Each one has the same purpose and design but is implemented multiple times
- Instead only use the error component and call that within the error.js and not-found.js files

fardinzam:
Screenshot 2023-10-25 at 9 35 10 PM

### DIFF
--- a/src/app/error.js
+++ b/src/app/error.js
@@ -1,17 +1,12 @@
 "use client";
+import Error from "@/components/dynamic/Error";
 
-export default function Error({ error }) {
+export default function InternalError() {
   return (
-    <div className="w-screen h-screen flex flex-col items-center justify-center fixed">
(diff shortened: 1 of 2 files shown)


src/app/error.js
```diff
@@ -1,17 +1,12 @@
...
-      </p>
-      <p className="text-center text-lg md:text-2xl font-bold text-black m-0">
-        Internal Server Error
-      </p>
-      <p className="text-center text-sm md:text-lg  text-hackathon-blue-200 m-0">
-        Please contact the web development team for assistance.
-      </p>
-    </div>
+    <Error
+      code={"500"}
```
shahdivyank: dont need curly braces, if its only strings

shahdivyank (approved):

PR #852, opened 2023-10-31
Server Checking Util

### LINKED ISSUE #803: Add Server Checking Util
- Add a new util function similar to authenticate
- Accepts object of attributes
- ensures that text inputs never exceed 100 characters
- ensures that base64 files/images are within 200mb
- Should be called on every api call to check no excessive data is being put in

vwong031:
- Closes #803


src/utils/checkServer.js
```diff
@@ -0,0 +1,21 @@
+export const checkServer = (data) => {
+  const errors = [];
+
+  for (const key in data.textInputs) {
```
shahdivyank: should be a foreach loop

src/utils/checkServer.js
```diff
@@ -0,0 +1,21 @@
...
+    if (data.textInputs[key].length > 100) {
+      errors.push("Text input ${key} exceeds 100 characters.");
+    }
+  }
+
+  for (const key in data.base64Files) {
```
shahdivyank: how does it differentiate between text and base64 files?

src/utils/validate.js
```diff
@@ -0,0 +1,18 @@
...
+        valid: false,
+      };
+    }
+  });
+
+  data.files.forEach((file, key) => {
+    if (file.length > 200 * 1024 * 1024) {
+      return { message: `File ${key} exceeds 200MB.`, valid: false };
+    }
+  });
```
shahdivyank, line 19: check for numbers as well

src/utils/validate.js
```diff
@@ -0,0 +1,18 @@
...
+      return {
+        message: `Text input ${key} exceeds 100 characters.`,
+        valid: false,
+      };
+    }
+  });
+
+  data.files.forEach((file, key) => {
+    if (file.length > 200 * 1024 * 1024) {
+      return { message: `File ${key} exceeds 200MB.`, valid: false };
```
shahdivyank: 200mb doesnt work for all of them 😭max is 800mb I think

shahdivyank:
Unfortunately we have to drop this feature cause server checking depends on the clauses provided by frontend and would make this too adaptive instead of the strict checking we were looking for :sob:

PR #854, opened 2023-11-01
added arrow key shortcuts

### LINKED ISSUE #829: Add Arrow Key Keyboard Shortcuts
image

- Pressing left or right arrow should change the week/month
- pressing a should trigger the "all" events filter

khandrew1:
* Added arrow key functions
* updated `UseEffect` to include view as well because it wasn't working without it cause of stuff with the key handler or something
* Closes #829


src/components/dynamic/admin/services/calendar/Events.jsx
```diff
@@ -25,6 +25,20 @@ const CalendarEvents = () => {
       case "w":
         setView("week");
         break;
+      case "ArrowRight":
+        if (view === "month") {
+          setDate(new Date(date.setMonth(date.getMonth() + 1)));
```
shahdivyank: should just be able to call event.NEXT, if you check the toolbar component you can see the way to change it

shahdivyank (approved):

PR #863, opened 2023-11-05
AgentEinstein/add error checking

### LINKED ISSUE #760: Add Error Checking
image

- If no team id is entered, it should reply with enter a valid team id

AgentEinstein:
image


src/components/dynamic/user/Team.jsx
```diff
@@ -37,6 +37,10 @@ const Team = ({ user, setUser }) => {
   };
 
   const handleJoin = () => {
+    if (id.team.length === 0) {
```
AgentEinstein: fixed

shahdivyank: u can just check for an empty string instead of character comparison

src/components/dynamic/user/Team.jsx
```diff
@@ -37,6 +37,10 @@ const Team = ({ user, setUser }) => {
   };
 
   const handleJoin = () => {
+    if (id.team == null) {
```
shahdivyank: triple equals and it should just be checking for empty string and not null

AgentEinstein: image
That doesn't seem to work. Instead, this jumps to the very last else statement on line 55.
image

AgentEinstein: On the other hand, !id.team seems to work, did you prefer that implementation instead?

shahdivyank: what are the initial values for the id.team?

AgentEinstein: it's probably empty

shahdivyank: check the values and do a type check for it instead of a general null check

shahdivyank (approved):

PR #875, opened 2023-11-09
Fuyuki/disable-reset-button

### LINKED ISSUE #710: Disable Reset Button When Applicable
image

- Reset button should be disabled if there are no judging rounds generated

Fuyuki325:
Fixed Disable Reset Button When Applicable #710

Disabled Reset1

### DIFF
--- a/src/components/dynamic/admin/services/judging/Toolbar.jsx
+++ b/src/components/dynamic/admin/services/judging/Toolbar.jsx
@@ -207,8 +207,9 @@ const Toolbar = ({ data, setData, judges, setJudges }) => {
             color="red"
             text="reset"
             onClick={handleReset}
-            disabled={data == 0}
+            disabled={data === null || []}
           />
+          {console.log(data)}
         </div>
         <div className="flex">
           {tags.map((tag, index) => (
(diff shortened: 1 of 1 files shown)


src/components/dynamic/admin/services/judging/Toolbar.jsx
```diff
@@ -203,7 +203,12 @@ const Toolbar = ({ data, setData, judges, setJudges }) => {
             <p className="mb-0 font-semibold mx-2"># of rotations</p>
             <Button color="green" text="generate" onClick={generate} />
           </form>
-          <Button color="red" text="reset" onClick={handleReset} />
+          <Button
+            color="red"
+            text="reset"
+            onClick={handleReset}
+            disabled={data == 0}
```
shahdivyank: do a type check (===)

Fuyuki325: type check (===) did not work so i did ==.

shahdivyank: type check means that the data is not a number, check the type of data and then compare it to that so that the type check works

shahdivyank: == is a jank equal and uses some assumptions, while === is way more strict

shahdivyank (approved):

PR #923, opened 2023-11-28
fixed calendar filters

### LINKED ISSUE #904: Fix Filtering
image
image

- Need to fix filtering, currently uses the old description formats, but it should use the new formats
- First hastag is the filterable group, and the second hashtag is the assignee

seanquiambao:
before: image

after:
image

### DIFF
--- a/src/components/dynamic/admin/services/calendar/Toolbar.jsx
+++ b/src/components/dynamic/admin/services/calendar/Toolbar.jsx
@@ -27,8 +27,8 @@ const CustomToolbar = ({
             event.description
               .split("\n")[0]
               .split("#")
-              .map((item) => item.trim())
               .filter((item) => item !== "")[0]
+              .trim()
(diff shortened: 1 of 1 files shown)


src/components/dynamic/admin/services/calendar/Toolbar.jsx
```diff
@@ -24,8 +24,12 @@ const CustomToolbar = ({
       setEvents(
         events.map((event) => {
           if (
-            event.description.split("\n")[1].split(": ")[1].toLowerCase() ===
-            value
+            event.description
+              .split("\n")[0]
+              .split("#")
+              .map((item) => item.trim())
+              .filter((item) => item !== "")[0]
```
shahdivyank, line 30: u can combine the trim and filters u can just do item.trim() !== ""

shahdivyank (approved):

PR #924, opened 2023-11-30
Prevent API Call

### LINKED ISSUE #745: Prevent API Call
image

- API call should not be triggered when there are no objects or when nothing has been selected
- Make sure the API call is not triggered

twang0323:


cypress/fixtures/admin.json
```diff
@@ -0,0 +1,135 @@
+[
```
shahdivyank: this file shouldnt be needed

src/components/dynamic/admin/dashboards/Toolbar.jsx
```diff
@@ -34,6 +34,16 @@ const Toolbar = ({
   const [toggle, setToggle] = useState(false);
 
   const onClick = (value) => {
+    if (!objects.some((obj) => obj.selected)) {
```
shahdivyank, line 37: this is duplicate code

shahdivyank (approved):

PR #970, opened 2023-12-24
consolidated forms routes

### LINKED ISSUE #946: Form Consolidation
Problem: Currently under the `src/app/form` there are a ton of folders/files which all just eventually call protected page and some component and the component itself has the code. The protected pages are also all the same same except for the page title

- Want to utilize dynamic routes to have a single `/form/[...type]` that would dynamically take the user to the correct route based on the value of type, checkout the documentation for more details
- This new route would call protected page and then render a component based on the provided type. 
- https://nextjs.org/docs/app/building-your-application/routing/dynamic-routes

Note:
- This opens up the possibility that someone enters something like `form/participant1234567890` which is invalid, but to code it will try to render something and might give them a blank screen, in situations like these we want to give them a 404 page

khandrew1:
* Consolidated forms by using dynamic routes
* Im just rendering the Error component instead of rerouting to a 404 because I was unsure of how to do that with the app router
* Closes #946


src/app/form/[type]/page.js
```diff
@@ -0,0 +1,41 @@
...
+    sponsor: <Sponsor />,
+    volunteer: <Volunteer />,
+  };
+
+  return components[route] ? (
+    <ProtectedPage title={`Form | ${route}`} restrictions={{}}>
+      {components[route]}
+    </ProtectedPage>
+  ) : (
+    <Error
```
Comment (shahdivyank) [inline src/app/form/[type]/page.js]: throw the error instead

shahdivyank (approved):

PR #981, opened 2023-12-27
consolidated admin

### LINKED ISSUE #947: Consolidate Admin
Problem: Currently under the `src/app/admin` there are a ton of folders/files which all just eventually call protected page and some component and the component itself has the code. The protected pages are also all the same same except for the page title

- Want to utilize dynamic routes to have a single `/admin/[...type]` that would dynamically take the user to the correct route based on the value of type, checkout the documentation for more details
- This new route would call protected page and then render a component based on the provided type. 
- https://nextjs.org/docs/app/building-your-application/routing/dynamic-routes

Note:
- This opens up the possibility that someone enters something like `admin/participant1234567890` which is invalid, but to code it will try to render something and might give them a blank screen, in situations like these we want to give them a 404 page

fardinzam:


fardinzam:
i missed the "committee: 1" tag for a few of them I'm gonna fix them a little later

shahdivyank (approved):

PR #983, opened 2023-12-29
made links apear conditionally

### LINKED ISSUE #910: Only Show Valid Links
image

- Only show the icon if the link exists, otherwise it just redirects to the same page
- Make sure these links open in new tabs

fardinzam:
Screenshot 2023-12-28 at 4 20 55 PM

### DIFF
--- a/src/components/dynamic/admin/services/judging/Table.jsx
+++ b/src/components/dynamic/admin/services/judging/Table.jsx
@@ -18,20 +18,18 @@ const Table = ({ data }) => {
                 <Tag color={COLORS["grayblue"]} text={name} />
                 <div className="flex justify-start w-full ml-2">
                   {links &&
-                    links.map((link, index) => {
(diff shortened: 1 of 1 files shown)


src/components/dynamic/admin/services/judging/Table.jsx
```diff
@@ -18,16 +18,20 @@ const Table = ({ data }) => {
...
-                      <Link
-                        key={index}
-                        href={link.link}
-                        target="_blank"
-                        className="m-0 p-0 text-black no-underline hover:!text-hackathon-blue-100 text-xl"
-                      >
-                        {ICONS[link.name]}
-                      </Link>
-                    ))}
+                    links.map((link, index) => {
```
shahdivyank: u can combine the map and if condition with a filter. do an implicit return

shahdivyank (approved):

PR #986, opened 2023-12-30
delete loading page

### LINKED ISSUE #959: Simplify
image

- We can just import and return the Loading component directly, there is no need to create a new component and then send that

seanquiambao:


shahdivyank:
We need to keep the page but just simplify it by importing the component and directly exporting the component again, we dont need to make a new one

PR #1002, opened 2024-01-02
fix accordion

mailynnguyen:
image


shahdivyank:
Manually integrated due to table migration

PR #1012, opened 2024-01-04
changed failed api message

### LINKED ISSUE #900: Throw Error on Failed API
- Currently the API silently passes a 500 error message, which usually results in a toast showing up that says Internal Server Error
- Instead we need to throw our custom error and give the user a big error page telling them that something is messed up, so it does not go unnoticed
- Make sure to throw the custom error that has been created

fardinzam:
Screenshot 2024-01-03 at 8 58 58 PM


fardinzam:
for now I only changed Admin because I want to confirm my changes are good before changing the other files

src/app/api/dashboard/[type]/route.js
```diff
@@ -58,7 +58,7 @@ export async function POST(req, { params }) {
     return res.json({ message: "OK" }, { status: 200 });
   } catch (err) {
     return res.json(
-      { message: `Internal Server Error: ${err}` },
+      { message: `Internal Server Error: ${err}, Stack trace: ${err.stack}` },
```
Comment (shahdivyank) [inline src/app/api/dashboard/[type]/route.js]: u can call this just stack

src/components/dynamic/form/Admin.jsx
```diff
@@ -19,27 +21,60 @@ const Admin = () => {
...
+            code: statusCode,
+            name: "Internal Server Error",
+            message: "Please reach out the dev team with the below message",
+            dev: responseBody.message,
+          });
+        }
+      })
+      .catch((error) => {
+        if (!errorObj) {
+          setErrorObj({
```
shahdivyank, line 45: this should be thrown instead, checkout the protected page, and how it implements throwing errors

src/utils/api.js
```diff
@@ -3,8 +3,20 @@ export const api = async ({ url, method, body }) => {
     method: method,
     body: JSON.stringify(body),
   });
-
   const data = await response.json();
 
   return data;
 };
+
+export const apiCallWithStatusCode = async ({ url, method, body }) => {
```
shahdivyank, line 5: u can modify the existing api call instead of making a new one, or u can also put a try catch on the api call and throw an error if goes into the catch

src/components/dynamic/form/Admin.jsx
fardinzam: the onSubmit is swallowing the error being thrown, it blocks it from the error message appearing, I found others with similar issues but I can't find anything on how to resolve it, otherwise I'd be able to throw 
https://github.com/redux-form/redux-form/issues/1859

src/utils/api.js
fardinzam: throwing here has the same problem as the comment above, I just made api use apiCallWithStatusCode

shahdivyank:
testing failed

src/app/api/dashboard/[type]/route.js
```diff
@@ -58,7 +58,7 @@ export async function POST(req, { params }) {
     return res.json({ message: "OK" }, { status: 200 });
   } catch (err) {
     return res.json(
-      { message: `Internal Server Error: ${err}` },
+      { message: `Internal Server Error: ${err}, Stack: ${err.stack}` },
```
Comment (shahdivyank) [inline src/app/api/dashboard/[type]/route.js:61]: stack should be lowercase

src/components/dynamic/form/Admin.jsx
```diff
@@ -4,12 +4,14 @@ import { useState } from "react";
 import Form from "@/components/dynamic/form/Form.jsx";
 import { FIELDS, ATTRIBUTES } from "../../../data/dynamic/form/Admins";
 import { useSession } from "next-auth/react";
-import { api } from "@/utils/api";
+import { apiCallWithStatusCode } from "@/utils/api";
```
shahdivyank, line 7: should just be called API

PR #1013, opened 2024-01-04
parallel read/write user and admin

### LINKED ISSUE #973: Parallel Read/Writes - User
https://stackoverflow.com/questions/58897274/what-is-the-fastest-way-to-write-a-lot-of-documents-to-firestore

- Currently, we implement sequential writes which is super slow since we wait for the previous write to finish before making another one
- We want to convert over to using parallel writes so we can make these operations much faster and less of a chance of failing or timing out

rfairooz:
I believe these are the 2 files I needed to change for user. Functionality is working properly I don't think anything is broken in the backend. I saw the data fetching time was much much faster. Let me know if this is okay.


src/app/api/dashboard/[type]/route.js
```diff
@@ -41,18 +41,22 @@ export async function POST(req, { params }) {
       ATTRIBUTES[params.type].forEach((attribute) => {
         element[attribute] = data[attribute];
       });
-      await updateDoc(doc(db, "users", user.id), {
-        ...element,
-        timestamp: Timestamp.now(),
-        [`roles.${params.type}`]: 0,
-      });
+      await Promise.all(
```
Comment (shahdivyank) [inline src/app/api/dashboard/[type]/route.js]: this odenst need a promise.all since its a single post

src/app/api/dashboard/[type]/route.js
```diff
@@ -41,18 +41,22 @@ export async function POST(req, { params }) {
...
+        })
+      );
 
-      await EmailService.send({
-        email: user.email,
-        id: process.env.EMAIL_CONFIRMATION_TEMPLATE,
-        name: user.name,
-        position: params.type.slice(0, -1).toUpperCase(),
-      });
+      await Promise.all(
```
Comment (shahdivyank) [inline src/app/api/dashboard/[type]/route.js]: dont need promise.all here

src/app/api/dashboard/[type]/route.js
```diff
@@ -85,21 +89,22 @@ export async function GET(req, { params }) {
           where(`roles.${params.type}`, "in", [-1, 0, 1])
         )
       );
-      snapshot.forEach((doc) => {
+      const promises = snapshot.docs.map(async (doc) => {
```
Comment (shahdivyank) [inline src/app/api/dashboard/[type]/route.js]: get doesnt need it either since its a single read operation. the for loop here is not something we will be able to optimize

src/app/api/dashboard/[type]/route.js
```diff
@@ -85,21 +89,22 @@ export async function GET(req, { params }) {
...
           ...element,
           uid: doc.id,
           timestamp: data.timestamp,
           status: data.roles[params.type],
           selected: false,
           hidden: false,
-        });
+        };
       });
+      output.push(...(await Promise.all(promises)));
```
Comment (shahdivyank) [inline src/app/api/dashboard/[type]/route.js]: need promise.all for put and delete

shahdivyank (approved):

PR #1018, opened 2024-01-05
Make stats page use statistics document instead of querying users

### LINKED ISSUE #949: Statistics Improvements
image

- Currently the statistics page runs multiple queries to get the various users and their counts
- We want to simplify this process so that we do something to how we get data for events
- We want to store everything in a statistics collection document that we can read, first column should read `statistics` and second column should also read `statistics`
- The last column should have information per user type
- We should keep track of how many users applied to various roles (need to add a database call in forms to update pending users)
- Then we also need to add database calls to admin portals when their status updates to accepted or rejected
- This reduces our entire API call to one document instead of multiple queries which take a very long time

pondohoo:
also changed it so it references when users are status "1" instead of 1 0 or -1


shahdivyank:
Link issue

src/app/api/dashboard/[type]/route.js
```diff
@@ -165,6 +181,11 @@ export async function DELETE(req, { params }) {
   try {
     if (types.has(params.type)) {
       objects.map(async (object) => {
+        const snapshot = await getDoc(doc(db, "users", object));
```
Comment (shahdivyank) [inline src/app/api/dashboard/[type]/route.js:184]: the object itself should have the role as well so we dont need to make an API call for it

src/app/api/statistics/route.js
```diff
@@ -1,14 +1,43 @@
...
-  where,
-  query,
-  collection,
-  getCountFromServer,
-  getDocs,
-} from "firebase/firestore";
+import { collection, getDocs, getDoc, doc } from "firebase/firestore";
 import { authenticate } from "@/utils/auth";
 
+// used for initial setup, in case becomes innacurate
```
shahdivyank: remove comments

src/app/api/statistics/route.js
```diff
@@ -23,62 +52,17 @@ export async function GET() {
...
-    const sponsors = (
-      await getCountFromServer(
-        query(
-          collection(db, "users"),
-          where("roles.sponsors", "in", [-1, 0, 1])
-        )
-      )
-    ).data().count;
+    // await syncStatsWithDatabase();
+    const docStats = await getDoc(doc(db, "statistics", "statistics"));
```
shahdivyank: you can destructure this:
```jsx
const { participants } = await getDoc(doc(db, "statistics", "statistics"));
```

src/components/dynamic/admin/services/statistics/Statistics.jsx
```diff
@@ -3,6 +3,7 @@ import Title from "@/components/dynamic/admin/Title";
 import { useEffect, useState } from "react";
 import Tabs from "./Tabs";
 import Loading from "@/components/dynamic/Loading";
+import { api } from "@/utils/api";
```
shahdivyank, line 6: remove if not being used?!

pondohoo: image

this is undefined without that include

shahdivyank (approved):

PR #1024, opened 2024-01-09
toastUtil

### LINKED ISSUE #1000: Make Toast Util
- Need to have stricter/more useful toasting
- Need to make a separate util that will handle toasting, ultimately we will move to TailwindCSS toasts
- Abstract all toasts to a util

Fuyuki325:
I have no idea if i did this right


Fuyuki325:
I made a toast utility but I never called it so I think that's why it never showed up in the commit.

src/utils/toast.js
```diff
@@ -0,0 +1,18 @@
+const makeToast = (message, type) => {
+  // Create toast element
```
shahdivyank: remove comments

src/utils/toast.js
```diff
@@ -0,0 +1,18 @@
...
+  const toastElement = document.createElement("div");
+  toastElement.textContent = message;
+  toastElement.className = `toast ${
+    type === "success" ? "bg-green-500 text-white" : "bg-red-500 text-white"
+  }`;
+
+  // append toast to doc
```
shahdivyank: we should still be using react-hot-toast behind the scenes. we just want to group all the toasts and have one entry point so we can easily make changes to our toasting

src/utils/toast.js
```diff
@@ -0,0 +1,18 @@
+const makeToast = (message, type) => {
```
shahdivyank: should just be called `toast`

Fuyuki325: Screenshot 2024-01-09 102542

If I use react-hot-toast and name the function toast, isn't there two things named toast then? I think it gave me that error because of that.

shahdivyank: call it Toaster

Fuyuki325: image
I am planning to just call it toaster instead of Toaster because of this error from ESLint. On it right now

Fuyuki325:
image
How the work currently looks. Maybe I should delete the redundant emojis?

src/app/user/join/[teamID]/page.js
```diff
@@ -18,10 +18,10 @@ export default function page({ params }) {
       body: { team: params.teamID },
     }).then((response) => {
       if (response.message !== "OK") {
-        toast(`❌ ${response.message}`);
+        toaster(`❌ ${response.message}`, "error");
         return;
       }
-      toast("✅ Successfully joined team!");
+      toaster("✅ Successfully joined team!", "success");
```
Comment (shahdivyank) [inline src/app/user/join/[teamID]/page.js]: icon should be implied by the type of toast

src/utils/toast.js
```diff
@@ -0,0 +1,22 @@
...
+  const options = {
+    style: {
+      borderRadius: "10px",
+      background: "#333",
+      color: "#fff",
+    },
+    className:
+      type === "success" ? "bg-green-500 text-white" : "bg-red-500 text-white",
+  };
+  if (type === "success") {
```
shahdivyank: switch case cause we will put in quite a few other types as well

src/utils/toast.js
```diff
@@ -0,0 +1,22 @@
+import { toast } from "react-hot-toast";
+
+const toaster = (message, type) => {
+  const options = {
```
shahdivyank: should all be in tailwind

src/app/user/join/[teamID]/page.js
Comment (Fuyuki325) [reply src/app/user/join/[teamID]/page.js]: Do I remove the icons?

Fuyuki325:
image


I successfully did it but I was unable to do it in tailwindCSS unfortunately. I did this but it wouldn't work
```
toast.success(message, {
      className: 'bg-green-500 text-white',
    });
```

Fuyuki325:
I reset my branch to be like dev, pulled from it, and force pushed to my remote branch.

src/utils/toaster.js
```diff
@@ -0,0 +1,31 @@
+import { toast } from "react-hot-toast";
+
+const toaster = (message, type) => {
+  switch (type) {
+    case "success":
+      toast.success(message, {
+        style: {
+          backgroundColor: "#34D399",
```
shahdivyank: use tailwind

Fuyuki325:
image
Screenshot 2024-02-06 204231

shahdivyank (approved):

PR #1063, opened 2024-01-16
Parallel Read/Write - Forms

### LINKED ISSUE #971: Parallel Read/Writes - Forms
https://stackoverflow.com/questions/58897274/what-is-the-fastest-way-to-write-a-lot-of-documents-to-firestore

- Currently, we implement sequential writes which is super slow since we wait for the previous write to finish before making another one
- We want to convert over to using parallel writes so we can make these operations much faster and less of a chance of failing or timing out

vwong031:
- Closes #971

### DIFF
--- a/src/app/api/dashboard/feedback/route.js
+++ b/src/app/api/dashboard/feedback/route.js
@@ -23,7 +23,6 @@ export async function POST(req) {
   }
 
   const {
-    rating,
     additionalComments,
     eventSource,
     improvements,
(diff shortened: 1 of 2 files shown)


src/app/api/dashboard/feedback/route.js
```diff
@@ -32,15 +32,17 @@ export async function POST(req) {
...
-    await addDoc(collection(db, "feedback"), {
-      rating: parseInt(rating),
-      additionalComments,
-      eventSource,
-      improvements,
-      notBeneficial,
-      helpful,
-      status: 0,
-    });
+    await Promise.all(
```
shahdivyank: this doesn't need promise.all since its only a one time api call

shahdivyank (approved):

PR #1065, opened 2024-01-17
added SelectAll component

### LINKED ISSUE #979: Reduce Checkboxes
- On any given form, we have a series of checkboxes that users need to click thru each one and accept
- We can simplify this to a single checkbox that says, they accept all of the above 
- This will require creating a new component for Terms and Conditions, since before it would just use the checkbox and map through it, we will need the same model where we have an array of conditions that we map through and a final checkbox at the bottom

fardinzam:
Screenshot 2024-01-16 at 8 28 25 PM


shahdivyank:
Failing build check

shahdivyank:
UI wise, see if you can make the text into a bullet point list

src/components/dynamic/form/SelectAll.jsx
```diff
@@ -0,0 +1,29 @@
+"use client";
+// import { usePathname } from 'next/navigation'
+import Checkbox from "../Checkbox";
+
+const SelectAll = ({ toggle, onClick, required, text }) => {
```
shahdivyank: shouldn't need a separate component for this, should be able to use a single checkbox at the end and map thru the items within the questions component

src/components/dynamic/form/form/Questions.jsx
```diff
@@ -118,6 +118,42 @@ const Questions = ({
               ))}
             </>
           )}
+          {field.input === "checkbox" && (
```
shahdivyank: this is something that can be hard coded right before the MLH code of conduct

src/components/dynamic/form/TermsConditions.jsx
```diff
@@ -0,0 +1,37 @@
...
+  toggle,
+  text,
+  onClick,
+  required,
+  color,
+}) => {
+  return (
+    <>
+      <p className="mb-1 mt-3 font-semibold">
+        {"Terms and Conditions"}
```
shahdivyank, line 15:
```suggestion
        "Terms and Conditions"
```

shahdivyank (approved):

PR #1073, opened 2024-01-21
fix judging team query

menthy-wu:


src/app/api/judging/route.js
```diff
@@ -50,6 +50,7 @@ export async function GET() {
         });
       }
     });
+    console.log(teams);
```
shahdivyank: remove console log

src/app/api/judging/route.js
```diff
@@ -28,12 +28,12 @@ export async function GET() {
 
   try {
     const teamsSnapshot = await getDocs(
-      query(collection(db, "teams"), where("status", "==", 1))
+      query(collection(db, "teams"), where("status", ">=", 0))
```
shahdivyank: we only want teams that are confirmed so == 1

shahdivyank:
Will deal with judging query after team submission is finalized

PR #1093, opened 2024-02-01
add phone number validation

### LINKED ISSUE #1061: Form Validation - Phone Number
- Currently forms suffer from the possibility of incorrect data (ie inputing a string when it expects a phone number)
- The first step in the form validation process is to implement validation on phone number ensuring that it has 10 total digits
- The formatting of the numbers should not matter (ie 123 456 7890 vs (123) 456 7890 or 1234567890), simply need to ensure that all characters are numbers
- Need to create a clean interface that can be adapted for all input fields
- Something along the lines of a `validationFn`

pondohoo:
onBlur changes field to this if invalid

image

if field contains "invalid", triggers the "Please complete all required fields" box

works on user dashboard

these are all the valid formats

(310)5554444
3105554444
310 555 4444
(310)555-4444
(310)-555-4444
310-435-3333
310435-3333
310-4353333
(310) 555 4444
(310) 5554444
(310)555 4444


src/data/dynamic/form/Judge.js
```diff
@@ -48,6 +48,8 @@ export const FIELDS = {
     maxLength: 50,
     width: 12,
     required: true,
+    regexValidation:
+      /^(\+\d{1,2}\s?)?1?\-?\.?\s?\(?\d{3}\)?[\s.-]?\d{3}[\s.-]?\d{4}$/,
```
shahdivyank: make a new regex file, since these regex statements are long and not as easy to read cause regex is hard to read, it can be very easy to mess this up across multiple forms

src/components/dynamic/user/User.jsx
```diff
@@ -14,13 +14,22 @@ import Button from "../Button";
...
 
 const User = ({ user, setUser, edit, setEdit }) => {
   const handleEdit = () => {
     setEdit(true);
   };
 
   const handleSave = async () => {
+    if (
+      Object.values(user).some(
+        (value) => typeof value === "string" && value.includes("Invalid")
```
shahdivyank, line 27: do u need the typeof value check?

shahdivyank (approved):

PR #1097, opened 2024-02-04
added default resume + teams participant

### LINKED ISSUE #1060: Add Default for Resume and Team
image

- The user is given the choice to upload a resume, if they do not upload resume there should be an N/A that shows up to indicate no resume
- The same concept should be applied to team

seanquiambao:
image

couldn't really figure out how to show that a participant is assigned to team, but it should show N/A if getValue() is undefined.

### DIFF
--- a/src/data/dynamic/admin/Participants.js
+++ b/src/data/dynamic/admin/Participants.js
@@ -45,7 +45,7 @@ export const COLUMNS = [
     accessorKey: "team",
     header: "Team",
     width: "w-3/12",
-    cell: ({ getValue }) => <div>{getValue() ? getValue() : "N/A"}</div>,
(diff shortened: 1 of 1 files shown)


src/data/dynamic/admin/Participants.js
```diff
@@ -45,16 +45,16 @@ export const COLUMNS = [
     accessorKey: "team",
     header: "Team",
     width: "w-3/12",
-    cell: ({ getValue }) => <div>{getValue()}</div>,
+    cell: ({ getValue }) => <div>{getValue() ? getValue() : "N/A"}</div>,
```
shahdivyank: u can do `getValue() ?? "N/A"`

shahdivyank (approved):

PR #1124, opened 2024-02-17
Consolidated Toaster into Singular Component

### LINKED ISSUE #1123: Consolidate Into a Single Component
image

- The only changing aspect is the icon, as a result, we can create a singular toast component that can be called inside of the toast.custom with props for text and icon

chonml:
Given that there is a lot of similarity between between each type of toast, I condensed it into a singular component that checks which status toaster receives. I just created a mapping and instead of reusing html code I used the mappings to assign icons to the different type props.

```javascript
const toaster = (message, type) => {
  const iconMapping = {
    success: <FaCheckCircle className="text-green-500" size="35px" />,
    error: <MdError className="text-red-500" size="35px" />,
  };

  return toast.custom(
    <div className="flex items-center justify-center bg-white pl-2 pr-5 md:pr-6 py-4 w-4/12 md:w-2/12 shadow-md rounded">
      {iconMapping[type] || <MdError className="text-red-500" size="35px" />}
      <div className="inline ps-4 text-sm font-normal text-black">
        {message}
      </div>
    </div>
  );
};
```

### DIFF
--- a/src/utils/toaster.js
+++ b/src/utils/toaster.js
@@ -5,35 +5,32 @@ import { FaCheckCircle } from "react-icons/fa";
 const toaster = (message, type) => {
   switch (type) {
     case "success":
-      toast.custom(
-        <div className="flex items-center justify-center bg-white pl-2 pr-5 md:pr-6 py-4 w-4/12 md:w-2/12 shadow-md rounded">
-          <FaCheckCircle className=" text-green-500" />
-          <div className="inline ps-4 text-sm font-normal text-black">
-            {message}
-          </div>
-        </div>
-      );
+      toast.custom(consolidateToaster(message, type));
       break;
     case "error":
-      toast.custom(
-        <div className="flex items-center justify-center bg-white pl-2 pr-5 md:pr-6 py-4 w-4/12 md:w-2/12 shadow-md rounded">
-          <MdError className=" text-red-500" size="35px" />
-          <div className="inline ps-4 text-sm font-normal text-black">
-            {message}
-          </div>
-        </div>
-      );
+      toast.custom(consolidateToaster(message, type));
       break;
     default:
-      toast.custom(
-        <div className="flex items-center justify-center bg-white pl-2 pr-5 md:pr-6 py-4 w-4/12 md:w-2/12 shadow-md rounded">
-          <MdError className=" text-red-500" size="35px" />
-          <div className="inline ps-4 text-sm font-normal text-black">
-            {message}
-          </div>
-        </div>
-      );
+      toast.custom(consolidateToaster(message, type));
+      break;
   }
+};
+
+const consolidateToaster = (message, type) => {
+  const icon =
+    type === "success" ? (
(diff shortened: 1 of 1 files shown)


src/utils/toaster.js
```diff
@@ -5,35 +5,32 @@ import { FaCheckCircle } from "react-icons/fa";
...
-          </div>
-        </div>
-      );
+      toast.custom(consolidateToaster(message, type));
+      break;
   }
 };
 
+const consolidateToaster = (message, type) => {
+  const icon =
```
shahdivyank: keep this a mapping so we can easily add more icons as needed

src/utils/toaster.js
```diff
@@ -5,35 +5,32 @@ import { FaCheckCircle } from "react-icons/fa";
...
     case "success":
-      toast.custom(
-        <div className="flex items-center justify-center bg-white pl-2 pr-5 md:pr-6 py-4 w-4/12 md:w-2/12 shadow-md rounded">
-          <FaCheckCircle className=" text-green-500" />
-          <div className="inline ps-4 text-sm font-normal text-black">
-            {message}
-          </div>
-        </div>
-      );
+      toast.custom(consolidateToaster(message, type));
```
shahdivyank: you can integrate the toast.custom into the Toaster function

src/utils/toaster.js
```diff
@@ -5,35 +5,32 @@ import { FaCheckCircle } from "react-icons/fa";
...
+      break;
   }
 };
 
+const consolidateToaster = (message, type) => {
```
shahdivyank: call it just Toaster

src/utils/toaster.js
```diff
@@ -3,37 +3,19 @@ import { MdError } from "react-icons/md";
...
-      );
-  }
+  const iconMapping = {
+    success: <FaCheckCircle className="text-green-500" size="35px" />,
+    error: <MdError className="text-red-500" size="35px" />,
+  };
+
+  return toast.custom(
+    <div className="flex items-center justify-center bg-white pl-2 pr-5 md:pr-6 py-4 w-4/12 md:w-2/12 shadow-md rounded">
+      {iconMapping[type] || <MdError className="text-red-500" size="35px" />}
```
shahdivyank: dont keep a default state, this is an internal thing we use, so can be super tight on usage

src/utils/toaster.js
```diff
@@ -3,37 +3,19 @@ import { MdError } from "react-icons/md";
...
-      toast.custom(
-        <div className="flex items-center justify-center bg-white pl-2 pr-5 md:pr-6 py-4 w-4/12 md:w-2/12 shadow-md rounded">
-          <MdError className=" text-red-500" size="35px" />
-          <div className="inline ps-4 text-sm font-normal text-black">
-            {message}
-          </div>
-        </div>
-      );
-  }
+  const iconMapping = {
```
shahdivyank: call it icons (consistent with some other icon mappings we have throughout the repo)

shahdivyank (approved):

PR #1170, opened 2024-03-05
Final Sync Template Fix

khandrew1:
* actually finally works
* runs at 6 am everyday
* constantly updates sync template PR to keep up to date

### DIFF
--- a/.github/workflows/sync-template.yaml
+++ b/.github/workflows/sync-template.yaml
@@ -1,7 +1,7 @@
 name: Sync Changes from Template
 on:
   schedule:
-    # Run at 6:00 AM every two days
+    # Run at 6:00 AM every day
     - cron: "0 14 * * *"
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/sync-template.yaml
```diff
@@ -1,8 +1,8 @@
 name: Sync Changes from Template
 on:
   schedule:
-    # Run at 6:00 AM every two days
-    - cron: "0 6 */2 * *"
+    # Run at 6:00 AM every day
```
shahdivyank, line 4:
```suggestion
    # Run at 6:00 PM every day
```

shahdivyank (approved):

.github/workflows/sync-template.yaml
```diff
@@ -1,8 +1,8 @@
 name: Sync Changes from Template
 on:
   schedule:
-    # Run at 6:00 AM every two days
-    - cron: "0 6 */2 * *"
+    # Run at 6:00 PM every day
```
shahdivyank:
```suggestion
    # Run at 6:00 AM every day
```

PR #1207, opened 2024-03-12
made it so that the url temporarily flash disabled

### LINKED ISSUE #1149: Move `disabled` State to URL Search Params
- https://nextjs.org/docs/app/api-reference/file-conventions/page#searchparams-optional
- https://github.com/shahdivyank/r-parts/pull/246/files

JohnZ9865:
Closes #1149 
Flashes true for an instant, then goes back to false, like how it was with useState.
image
image

### DIFF
--- a/src/components/dynamic/admin/services/contacts/Contact.jsx
+++ b/src/components/dynamic/admin/services/contacts/Contact.jsx
@@ -5,6 +5,7 @@ import Select from "@/components/dynamic/Select";
 import { useState } from "react";
 import { api } from "@/utils/api";
 import toaster from "@/utils/toaster";
+import { useSearchParams, usePathname, useRouter } from "next/navigation";
 
 const STATUSES = ["confirmed", "pending", "not attending"];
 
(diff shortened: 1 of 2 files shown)


src/components/dynamic/admin/services/contacts/Contact.jsx
```diff
@@ -5,6 +5,7 @@ import Select from "@/components/dynamic/Select";
 import { useState } from "react";
 import { api } from "@/utils/api";
 import toaster from "@/utils/toaster";
+import { useSearchParams, usePathname, useRouter } from "next/navigation";
```
shahdivyank, line 8: it should use server side params https://nextjs.org/docs/app/api-reference/file-conventions/page#searchparams-optional

src/components/dynamic/admin/services/contacts/Contact.jsx
```diff
@@ -5,6 +5,7 @@ import Select from "@/components/dynamic/Select";
 import { useState } from "react";
```
shahdivyank, line 5: should not need usestate

PR #1216, opened 2024-03-14
made links for website and devpost open in new tab

### LINKED ISSUE #1178: Open DevPost and Website In New Tab
image

- Open devpost and website (external links) in new browser tabs

atharvanev:


atharvanev:
I didn't add a screenshot since I didn't know how I would show it opens in a new Tab

atharvanev:
also there is an pre-existing error from tab.onlick() in Navigation.jsx from before where every element doesn't have an "onclick: " so it causes an error, but I didn't  know if I should handle it, since the functionality is still there

shahdivyank (approved):

PR #1271, opened 2024-04-07
judging, rule, hackpack

### LINKED ISSUE #1249: Judging + Rules + Hackpacks
image

- Should be collapsable
- rules and judging will be bullet point items

menthy-wu:
image
image

### DIFF
--- a/src/components/user/Collapse.jsx
+++ b/src/components/user/Collapse.jsx
@@ -1,0 +1,25 @@
+import { RiArrowDownSLine } from "react-icons/ri";
+const Collapse = ({ text, children, setExpand, expand }) => {
+  return (
+    <div className="w-full bg-white rounded-xl p-4 drop-shadow-md">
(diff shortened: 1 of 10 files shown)


src/components/user/Collapse.jsx
```diff
@@ -0,0 +1,25 @@
+import { RiArrowDownSLine } from "react-icons/ri";
+const Collapse = ({ text, children, setExpand, expand }) => {
+  return (
+    <div className="w-full bg-white rounded-xl p-4 drop-shadow-md">
+      <div
+        className="flex items-center justify-between cursor-pointer text-xl font-bold"
+        onClick={() => setExpand(text === expand ? null : text)}
```
shahdivyank: it should not be null dependent, should be true/false or text and empty string

shahdivyank (approved):

PR #1279, opened 2024-04-12
Made jsx-a11y/alt-text required in eslint

### LINKED ISSUE #1278: Make `jsx-a11y/alt-text` Required in EsLint
There is a particular rule that is currently a warning, but we would like to address it immediately since its a simple fix, the rule is `jsx-a11y/alt-text`

RiyanaD:
Added the jsx-a11y/alt-text rule so that ESLint will report an error if it finds any img elements without an alt attribute.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -9,12 +9,15 @@
     "ecmaVersion": "latest",
     "sourceType": "module"
   },
-  "plugins": ["react","jsx-a11y"],
+  "plugins": ["react"],
   "rules": {
     "require-jsdoc": 0,
-    "jsx-a11y/alt-text": ["error", {
-      "elements": ["img"]
-    }]
+    "jsx-a11y/alt-text": [
+      1,
+      {
+        "elements": ["img"]
+      }
+    ]
   },
   "settings": {
     "react": {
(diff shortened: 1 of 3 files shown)


.eslintrc.json
```diff
@@ -9,9 +9,12 @@
     "ecmaVersion": "latest",
     "sourceType": "module"
   },
-  "plugins": ["react"],
+  "plugins": ["react","jsx-a11y"],
```
shahdivyank: this shouldn't be a plugin, just a rule

.eslintrc.json
```diff
@@ -9,9 +9,12 @@
...
   "rules": {
-    "require-jsdoc": 0
+    "require-jsdoc": 0,
+    "jsx-a11y/alt-text": ["error", {
```
shahdivyank: should just be 1 for error

package.json
```diff
@@ -54,6 +54,7 @@
     "eslint-config-google": "^0.14.0",
     "eslint-config-next": "^13.5.6",
     "eslint-config-prettier": "^9.0.0",
+    "eslint-plugin-jsx-a11y": "^6.8.0",
```
shahdivyank: we shouldn't need an additional plugin

shahdivyank (approved):

Issue #1316, opened 2024-05-03
Fix Weird Animation


shahdivyank:
<img width="453" alt="image" src="https://github.com/acm-ucr/hackathon-website/assets/36285610/cf297134-49d3-4f2e-9776-55525604fd35">

- The animation delay appears to be a bug and slow
- Make it smoother or remove the animation entirely

PR #1370, opened 2024-05-27
Finished weird animation fixbut had to implement it in a non tailwind way

arnavswami:
Not sure how to make it in a tailwind format and tried doing so. Was wondering how that would be done.

https://github.com/acm-ucr/hackathon-website/assets/68790299/75660e9d-ab01-470f-9da1-b30658331e80


arnavswami:
#1316 Fix Weird Animation

shahdivyank (approved):

PR #1442, opened 2024-06-06
Update lead form description

### LINKED ISSUE #1357: Fix Description
image

- Description needs to be fixed since the hackathon has already passed when the applicants are filling out the form

jonathanle03:
image

### DIFF
--- a/src/data/form/Leads.js
+++ b/src/data/form/Leads.js
@@ -9,7 +9,7 @@ export const FIELDS = {
     width: 12,
     texts: [
       `Welcome to ${data.name} ${
-        parseInt(data.year) + 1
+        data.date.getFullYear() + 1
       }. Thank you for considering to become a lead, we appreciate your enthusiasm to join ${
         data.name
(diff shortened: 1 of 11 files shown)


src/data/form/Leads.js
```diff
@@ -8,7 +8,11 @@ export const FIELDS = {
     input: "description",
     width: 12,
     texts: [
-      `Welcome to ${data.name}. Thank you for considering to become a lead, we appreciate your enthusiasm to join ${data.name}. ${data.name} is a ${data.description} hackathon spanning ${data.length} on ${data.date}.`,
+      `Welcome to ${data.name} ${
+        parseInt(data.year) + 1
```
shahdivyank: this should be based on the date attribute

jonathanle03: Is the the date attribute supposed to be a string or a date? If it's supposed to be a string, how would I get/change only the year part? If it's supposed to be a date, can I change the value to a random date?

shahdivyank (approved):

PR #1455, opened 2024-06-24
Feedback Parallel

### LINKED ISSUE #1379: Make Parallel Calls
- Within the feedback API, there are 3 Firestore calls and a Resend Email call that can be reduced/ran in parallel
- If the Firestore calls can run in parallel, then run them in parallel

seanquiambao:
Parallelized Firestore calls in GET DELETE and PUT requests. I did not parallelize getDocs and getCountFromServer, I think they're single operation?? Functionality works the same.

Let me know if I should fix anything!

### DIFF
--- a/src/app/api/dashboard/feedback/route.js
+++ b/src/app/api/dashboard/feedback/route.js
@@ -76,7 +76,6 @@ export const GET = async (req) => {
   }
 
   const output = [];
-
   try {
     let snapshot;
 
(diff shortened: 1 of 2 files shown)


src/app/api/dashboard/feedback/route.js
```diff
@@ -114,27 +113,30 @@ export const GET = async (req) => {
...
-        additionalComments,
-        eventSource,
-        improvements,
-        notBeneficial,
-        helpful,
-        status,
-      });
-    });
+
+    await Promise.all(
```
shahdivyank: promise.all on a map won't have any performance benefits

shahdivyank: There isn't much of an optimization we can make on the GET call since its already optimized via the query

shahdivyank (approved):

PR #1505, opened 2024-06-30
add Teams/Judges title

### LINKED ISSUE #1477: Add Section Title
image

- Since the current view is for teams and the secondary view is for judges, add a section header that illustrates that, the title should be on the upper left corner right before the input fields

mailynnguyen:
image
image

Two section titles "Judges" and "Teams", initial when loading page is "Teams", after pressing view button, title will switch to the other, alternating back and forth each time pressing the view button.
I think this was the assignment, hopefully, I understood it right.


src/components/admin/services/judging/Judging.jsx
```diff
@@ -22,6 +22,7 @@ const Judging = () => {
         judgesView={judges}
         setJudgesView={setJudges}
       />
+      {view ? <Title title="Judges" /> : <Title title="Teams" />}
```
shahdivyank: instead of having 2 components u can simplify down to changing the string in the component
```suggestion
      <Title title={view ? "Judges" : "Teams"} />
```

shahdivyank (approved):

PR #1511, opened 2024-07-02
made parallel call for checkin route

### LINKED ISSUE #1377: Make Parallel Calls
image

- Currently called sequentially when there is no dependence between the updateDoc and getDoc
- Make the updateDoc and getDoc parallel calls

Howardwheeler:
made parallel call for checkin route

### DIFF
--- a/src/app/api/checkin/route.js
+++ b/src/app/api/checkin/route.js
@@ -51,7 +51,8 @@ export const PUT = async (req) => {
   const { uid, event, name } = await req.json();
 
   try {
-    const [data] = await Promise.all([
+    // eslint-disable-next-line no-unused-vars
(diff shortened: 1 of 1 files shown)


src/app/api/checkin/route.js
```diff
@@ -51,11 +51,13 @@ export const PUT = async (req) => {
   const { uid, event, name } = await req.json();
 
   try {
-    await updateDoc(doc(db, "users", uid), {
-      events: arrayUnion(event),
-    });
+    const [data] = await Promise.all([
```
shahdivyank:
```suggestion
    const [_, data] = await Promise.all([
```

shahdivyank (approved):

PR #1518, opened 2024-07-06
User Portal Schedule

### LINKED ISSUE #1460: Implement Schedule Events Component
image

- Given events array similar to the calendar/live schedule pages, render out boxes for the various events
- Should take in an array as a prop so the toolbar can change the events array

stormyy00:
Screenshot 2024-07-07 at 1 41 18 AM


src/components/user/schedule/Events.jsx
```diff
@@ -1,5 +1,35 @@
-const Events = () => {
-  return <div>Events</div>;
+import React from "react";
+
+const Events = ({ event, setEvents }) => {
+  setEvents(null);
```
shahdivyank: this shouldn't be called on every component render

shahdivyank (approved):

PR #1528, opened 2024-07-11
added download button, need to implement download

### LINKED ISSUE #1372: Add Download
image

- Next to the view button, we need a download button that would download the file (either pdf or image) to the users local computer
- Use the download icon from react-icons
- Show a confirmation message similar to delete
- This will download all the entries. As a results, it will need to make api calls to the entire table instead of a paginated call
- https://www.npmjs.com/package/react-json-to-csv

fardinzam:
just need to implement download feature
Screenshot 2024-07-11 at 1 29 27 PM


src/components/admin/Tag.jsx
```diff
@@ -1,4 +1,4 @@
-const Tag = ({ color, text, onClick, classes, statuses }) => {
+const Tag = ({ color = {}, text, onClick, classes, statuses }) => {
```
shahdivyank: this is crazy dangerous, surely there is a better way to handle this

src/data/admin/Judges.js
```diff
@@ -53,7 +53,7 @@ export const COLUMNS = [
   {
     accessorKey: "photo",
     header: "Photo",
-    width: "w-1/12",
+    width: "w-[10%]",
```
shahdivyank: all the columns should equal to 100%

src/utils/utils.js
```diff
@@ -2,3 +2,24 @@ import { clsx } from "clsx";
 import { twMerge } from "tailwind-merge";
 
 export const cn = (...inputs) => twMerge(clsx(inputs));
+
+export const performDownload = async (src) => {
```
shahdivyank: this should be a download.js file inside the utils folder

shahdivyank (approved):

PR #1563, opened 2024-07-18
Split name into first and last name

### LINKED ISSUE #1558: Split into First Name and Last Name
image

- Currently the name is stored entirely as a single string from Google OAuth
- Need to split it into first name and last name based on MLH requirements
- Need to just visually display first and last name
- Should be an auto populated field
- You can simply grab the first word as the first name and the rest for the last name or flip the split and only the last word as last name and everything else is first name
- At the end the parsing should be ignored within the forms within the DB, this is purely a UI change

carlahauu:
image


src/components/form/Admin.jsx
```diff
@@ -10,9 +10,20 @@ import { STATUSES } from "@/data/admin/Admins";
 
 const Admin = () => {
   const { data: session } = useSession();
+
+  function extractFirstName(str) {
```
shahdivyank, line 14: 1. arrow function
2. put this as a util and have 1 function return both the first and last name and use that everywhere

src/components/form/Lead.jsx
```diff
@@ -10,11 +10,21 @@ import { STATUSES } from "@/data/admin/Leads";
...
+
+  function extractLastName(str) {
+    return str.trim().split(" ")[1];
+  }
+
   const [lead, setLead] = useState({
     ...ATTRIBUTES,
     name: session.user.name,
-    roles: session.user.roles,
+    firstName: extractFirstName(session.user.name),
```
shahdivyank, line 24: call the variables just first and last

src/components/form/Volunteer.jsx
```diff
@@ -11,9 +11,19 @@ import { STATUSES } from "@/data/admin/Volunteers.js";
...
 
+  function extractFirstName(str) {
+    return str.trim().split(" ")[0];
+  }
+
+  function extractLastName(str) {
+    return str.trim().split(" ")[1];
+  }
+
   const [volunteer, setVolunteer] = useState({
```
shahdivyank, line 22: for each form, when we submit, we need to remove the first and last separate and merge it back into just name so the DB remains the same

PR #1645, opened 2024-07-21
implemented shirts sizes to stats

### LINKED ISSUE #1577: Track Shirt Size + Display on Statistics
- One common query that we have to run every time is getting the shirt sizes across the various roles to get insights for preordering and allocating enough for judges
- Since we collect shirt sizes already on the forms, its just a matter of running the query, which is not viable especially since it has a SWE lead dependency
- We want to eventually put it on the Statistics page so that anyone with access can get the shirt data quickly.
- For every role, we need to store the shirt data (frequency of each size)
- This is to be displayed in a "Shirts" section on the statistics pages with charts (most likely similar to the ones of "Registrants") to showcase the shirt size spread across the various roles

stormyy00:
Screenshot 2024-08-09 at 2 21 57 PM


src/components/admin/services/statistics/Chart.jsx
```diff
@@ -12,15 +12,16 @@ import {
 } from "@/components/ui/chart";
 
 const Chart = ({ title, data }) => {
+  console.log(title, data);
```
shahdivyank: remove console log

stormyy00: LOLOL yeah

src/components/ui/.chart.jsx.swp
shahdivyank, line 1: this shouldn't be here?

src/app/api/statistics/route.js
```diff
@@ -18,11 +18,23 @@ export const GET = async () => {
   }
 
   try {
-    const [statistics, events] = await Promise.all([
+    const roles = [
```
shahdivyank: this shouldnt be a query on the entire table, it should be a form based update, check how the status gets updated

src/app/api/dashboard/[type]/route.js
```diff
@@ -227,16 +228,21 @@ export const PUT = async (req, { params }) => {
             preview: preview,
           });
 
+          const size = object.shirt;
+          console.log("PUT: ", size);
```
Comment (shahdivyank) [inline src/app/api/dashboard/[type]/route.js]: remove console log statements

src/components/admin/services/statistics/Chart.jsx
```diff
@@ -13,14 +13,14 @@ import {
 
 const Chart = ({ title, data }) => {
   const chartData = Object.entries(data).map(([type, value]) => ({
-    type: ROLES[type].label,
+    type: ROLES[type]?.label || SIZES[type]?.label,
     value: value,
-    className: ROLES[type].className,
-    fill: ROLES[type].fill,
+    className: ROLES[type]?.className || SIZES[type]?.className,
+    fill: ROLES[type]?.fill || SIZES[type]?.fill,
```
shahdivyank: make a better way so that we dont have to check if the value exists before calling fill/label on it

src/components/admin/services/statistics/Statistics.jsx
```diff
@@ -9,7 +9,7 @@ import Charts from "./Charts";
 
 const Statistics = () => {
   const [counts, setCounts] = useState(null);
-
+  console.log(counts);
```
shahdivyank: remove console log

shahdivyank (approved):

PR #1654, opened 2024-07-25
Converted all react-icons to lucide-react

### LINKED ISSUE #1549: Move All Icons to Lucide-React
- Currently we use react-icons for our icons
- Due to Shadcn migration, it installs Lucide for its icons and we will make the jump ship to Lucide
- There is a smaller family but it should cover all of our bases and synchronize our icons since we use various different families

sunbrandon:


sunbrandon:
Unable to convert Hackpacks.js to lucide icons since there is no lucide alternative for the react icons used

shahdivyank (approved):

PR #1665, opened 2024-07-29
releases js to ts

### LINKED ISSUE #1644: TypeScript: Releases
image

Derrick-Mao:


src/data/Releases.ts
```diff
@@ -1,4 +1,4 @@
-const RELEASES = {
+const RELEASES: { [key: string]: Date | { START: Date; END: Date } } = {
```
shahdivyank: create a separate type for this

Derrick-Mao: type separated

src/data/Releases.ts
```diff
@@ -1,4 +1,6 @@
-const RELEASES = {
+type ReleaseDate = Date | { START: Date; END: Date };
+
+const RELEASES: { [key: string]: ReleaseDate } = {
```
shahdivyank: type here should just be `RELEASES: release` everything else should be in the external type

shahdivyank (approved):

PR #1666, opened 2024-07-30
Typescript: Toaster Util

### LINKED ISSUE #1596: TypeScript: Toaster Util
image

nidheesh-m-vakharia:
I had to make a custom mapping as the LSP was showing an error that the ofunction toaster[type] returns a type any
image


src/utils/toaster.ts
```diff
@@ -0,0 +1,15 @@
+import { toast, ToastType } from "react-hot-toast";
+
+const toastMap: Record<ToastType, (message: string | JSX.Element) => string> = {
```
shahdivyank: it shouldn't be a union, the message will always be a string

shahdivyank: what is the Record for?

nidheesh-m-vakharia: > it shouldn't be a union, the message will always be a string

I added it just incase, i will change it to justs string

nidheesh-m-vakharia: > what is the Record for?

I used it because the handler function outputed from toast[type] did not really have a specific type, so i used the record to separate out the different handler fucntions

shahdivyank (approved):

PR #1698, opened 2024-08-08
Radio Shadcn

### LINKED ISSUE #1538: Integrate Radio with Shadcn
Requires https://github.com/acm-ucr/hackathon-website/issues/1537 to be complete

rfairooz:
Screenshot 2024-08-08 at 12 44 19 AM


src/components/Radio.jsx
```diff
@@ -20,28 +22,28 @@ const Radio = ({
         </div>
       )}
       {editable && (
-        <div className="grid w-full grid-cols-2 md:grid-cols-3">
+        <RadioGroup value={user[field]}>
```
shahdivyank, line 25: this value is dangerous, it should be abstracted out :sob:

src/components/Radio.jsx
```diff
@@ -20,28 +22,28 @@ const Radio = ({
...
-                    user[field] === option
-                      ? "bg-hackathon-green-300"
-                      : "bg-transparent"
-                  }`}
-                />
-              </div>
+              <RadioGroupItem
+                value={option}
+                defaultValue={option}
+                onClick={() => onClick(option, field)}
```
shahdivyank, line 35: onclick function should be handled in parent

src/components/Radio.jsx
```diff
@@ -1,3 +1,5 @@
+import { RadioGroup, RadioGroupItem } from "./ui/radio";
+
 const Radio = ({
```
shahdivyank, line 3: this file as a whole shouldn't be needed and should be removed, it creates an extra middle man for the radio component, all logic should be handled by the parent, while the child only handles the rendering

PR #1702, opened 2024-08-09
Made to Typescript

### LINKED ISSUE #1606: TypeScript: Calendar Service
image

annlee1226:
Screenshot 2024-08-08 at 6 57 45 PM


annlee1226:
Instead of having both the interfaces types and label, could we replace it with?:  
interface labels {[key: string]: {
    color: string;
    background: string;
    type: string;
  }
};

shahdivyank (approved):

PR #1752, opened 2024-08-17
Terms

### LINKED ISSUE #1730: TypeScript: Terms
image

Rama-Chunduri:
Screenshot 2024-08-17 at 7 52 40 AM


src/components/form/form/Terms.tsx
```diff
@@ -1,6 +1,12 @@
 import Checkbox from "@/components/Checkbox";
 
-const Terms = ({ options, toggle, onClick }) => {
+type terms = {
+  options: string[];
+  toggle: boolean;
+  onClick: () => {};
```
shahdivyank, line 6: () => void

needs to be void since it doesnt return anything

shahdivyank: checkout the other files for the format, it should be props and typed directly, dont need the react.fc

PR #1756, opened 2024-08-19
changes pg.js to ts

### LINKED ISSUE #1746: TypeScript: Hackathons
image

- do a simple import and export, do not create a new component

Shadow709:
changed hack js to ts


src/app/engineering/hackathons/page.tsx
```diff
@@ -1,10 +1,10 @@
 import Hackathons from "@/components/engineering/hackathons/Hackathons";
```
shahdivyank, line 1: this should also be similar to the biog page where it just imports and exports the component, no need for an extra div

shahdivyank (approved):

PR #1768, opened 2024-08-24
Changed event.jsx to tsx

### LINKED ISSUE #1733: TypeScript: Event
image

Shadow709:
eventTypeScriptFromJava

### DIFF
--- a/src/components/admin/services/calendar/Event.tsx
+++ b/src/components/admin/services/calendar/Event.tsx
@@ -1,4 +1,4 @@
-import React from "react"
+import React from "react";
 
 interface EventProps {
   event: {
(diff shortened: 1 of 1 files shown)


src/components/admin/services/calendar/Event.tsx
```diff
@@ -1,4 +1,14 @@
-const Event = ({ event, view }) => {
+import React from "react";
+
+interface EventProps {
+  event: {
+    start: Date & string;
```
shahdivyank: shouldn't be an & it should either be a string or a date not both

shahdivyank (approved):

PR #1773, opened 2024-08-26
adding stats for shirts, diet, schools 

### LINKED ISSUE #1578: Track School + Display School
- A common query that is usually manually run across the database is how many UCR attendants vs other schools
- Need to track which users are from which schools so that we can easily put that info on the statistics page
- Need to track similar to status 
- Track only for the participants
- Show in a "Participants" section on the statistics page
- Make sure to list out the other types of schools as well 
- Should ideally be displayed as a piechart ? but open to other suggestions (shadcn charts only)

### LINKED ISSUE #1579: Track Diet + Display Diet
- A common query that is usually manually run across the database is how many diet restrictions we have
- Need to track which users with diet restrictions so that we can easily put that info on the statistics page
- Need to track similar to status 
- Track for all users
- Show in a "Diet Restrictions" section on the statistics page
- Should ideally be displayed as a piechart per role? but open to other suggestions (shadcn charts only)
- Combinations can be dealt with uniquely (ie Vegan and Halal is different from Halal) (order the restrictions to get a consistent order)

stormyy00:
Screenshot 2024-09-13 at 2 25 34 AM

added inner ring using ```<Pie />``` component to visualize the status for each chart to make it easier to gather and find information due to the amount of charts present


src/app/api/dashboard/[type]/route.js
```diff
@@ -205,6 +207,8 @@ export const PUT = async (req, { params }) => {
           await updateDoc(doc(db, "users", object.uid), {
             [`roles.${params.type}`]: status,
             [`roles.${params.type}`]: object.shirt,
+            [`roles.${params.type}`]: object.diet,
```
Comment (shahdivyank) [inline src/app/api/dashboard/[type]/route.js]: roles should not be concerned with statistics

shahdivyank:
Keep different pie charts, integrating them makes them more difficult to look at :sob:

src/components/admin/services/statistics/Chart.jsx
```diff
@@ -11,8 +11,23 @@ import {
   ChartTooltipContent,
 } from "@/components/ui/chart";
 
-const Chart = ({ title, data }) => {
-  const ITEMS = { ...ROLES, ...SIZES };
+const Chart = ({ title, status = null, data }) => {
+  console.log("status", status);
```
shahdivyank: remove console logs

src/components/admin/services/statistics/Charts.jsx
```diff
@@ -1,19 +1,52 @@
...
+                          <Chart
+                            key={index}
+                            status={type}
+                            title={category}
+                            data={data}
+                          />
+                        );
+                      });
+                  } else if (key === "status") {
+                    if (Object.values(data).some((count) => count > 0)) {
```
shahdivyank: combine with the else if

shahdivyank (approved):

PR #1804, opened 2024-09-06
Add Subheadings

### LINKED ISSUE #1576: Add Subheadings
image

- Need to have subsections, which show "Registrations" and "Attendance"

nidheesh-m-vakharia:
image

### DIFF
--- a/src/components/admin/Subtitle.tsx
+++ b/src/components/admin/Subtitle.tsx
@@ -1,8 +1,4 @@
-type props = {
-  title: string;
-};
-
-const Subtitle = ({ title }: props) => {
+const Subtitle = ({ title }: { title: string }) => {
   return (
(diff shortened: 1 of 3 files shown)


src/components/admin/Subtitle.tsx
```diff
@@ -0,0 +1,16 @@
+type props = {
```
shahdivyank: should be within in the statistics folder

src/components/admin/Subtitle.tsx
```diff
@@ -0,0 +1,12 @@
+const Subtitle = ({ title }: { title: string }) => {
```
shahdivyank: extract the types

shahdivyank (approved):

PR #1808, opened 2024-09-07
toolbar filtering by description tags

### LINKED ISSUE #1682: Fix Toolbar Filtering
image

- Should be able to filter
- Client side filtering only

FelixMDM:
toolbarfilteringhackweek
toolbarfilteringhackathon


src/components/user/schedule/Schedule.jsx
```diff
@@ -10,10 +10,20 @@ const Schedule = ({ eventList }) => {
...
-      setFilteredEvents(events);
+    if (filterType === "hackweek") {
+      setFilteredEvents(
+        events.filter((event) => event.description.includes("#workshop")),
+      );
+    } else if (filterType === "hackathon") {
+      setFilteredEvents(
+        events.filter(
+          (event) =>
+            event.description.includes("#general") ||
```
shahdivyank: there can be workshops during the hackathon as well, the filtering should be based on the days currently being looked at, so for hackathon it should only be Saturday and sunday, while hackweek looks at mon - fri

shahdivyank (approved):

PR #1810, opened 2024-09-10
added input otp for number of rotations

### LINKED ISSUE #1678: Used OTP Input
image

- Use OPT Input from Shadcn for the number of rounds, allow for 2 digits

1rwen:
resolve #1678 
image

### DIFF
--- a/src/components/admin/services/judging/Toolbar.jsx
+++ b/src/components/admin/services/judging/Toolbar.jsx
@@ -8,6 +8,11 @@ import Popup from "../../Popup";
 import toaster from "@/utils/toaster";
 import { useRouter } from "next/navigation";
 import { api } from "@/utils/api";
+import {
+  InputOTP,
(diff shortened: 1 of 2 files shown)


src/components/admin/services/judging/Toolbar.jsx
```diff
@@ -203,15 +208,12 @@ const Toolbar = ({
...
-          <Input
-            setObject={setInput}
-            object={input}
-            label="rotations"
-            showLabel={false}
-            maxLength={2}
-            placeholder="ie. 5"
-            clear={true}
-          />
+          <InputOTP maxLength={2}>
```
shahdivyank: needs to handle the logic as well, currently these numbers only get visually seen but not tracked. You can check if the numbers are being tracking by console logging their values in the functions where they are supposed to be called

shahdivyank (approved):

PR #1811, opened 2024-09-11
changed length datatype to number

### LINKED ISSUE #1585: Change Datatype
image

- Should be of a Number type, since all hackathons are described in hours

chairfu:
image

### DIFF
--- a/src/data/Config.ts
+++ b/src/data/Config.ts
@@ -20,7 +20,7 @@ const data: config = {
   short_name: "Hackathon Site",
   email: "contact.acmucr@gmail.com",
   description: "DESCRIPTION",
-  length: 0,
+  length: 24,
   date: new Date("2024-04-12T13:20:00"),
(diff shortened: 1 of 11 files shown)


src/data/Config.ts
```diff
@@ -20,7 +20,7 @@ const data: config = {
   short_name: "Hackathon Site",
   email: "contact.acmucr@gmail.com",
   description: "DESCRIPTION",
-  length: "HACKATHON_LENGTH",
+  length: 0,
```
shahdivyank:
```suggestion
  length: 24,
```

shahdivyank:
check if we need to append hours wherever this variable is used. we expect 24 hours to show up, but since length is a number the strings need to handle the hours part

chairfu:
image
^added "hours" to all the forms and ideas.js as shown above (those were the only files that used data.length)
image

shahdivyank (approved):

PR #1819, opened 2024-09-13
add next-route-visualizer

stormyy00:
image


src/app/visualize/page.tsx
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+import Visualizer from "next-route-visualizer";
```
shahdivyank, line 1: return just the component

should just be 2 lines, one for the import and one for the default export

shahdivyank (approved):

shahdivyank (approved):

PR #1845, opened 2024-09-16
Shadow/cardlink

### LINKED ISSUE #1826: Enable Click on Card
image

- Clicking on the entire card should be able to open the hackpack instead of just the title

Shadow709:
card

### DIFF
--- a/src/components/user/resources/Hackpack.tsx
+++ b/src/components/user/resources/Hackpack.tsx
@@ -11,39 +11,37 @@ type props = {
 
 const Hackpack = ({ text, techs, link, description }: props) => {
   return (
-    <div className="rounded-xl bg-white p-3">
-      <Link
-        data-cy="hackpack-link"
-        className="items-center justify-between opacity-100 hover:opacity-70"
-        href={link}
-        target="_black"
+    <Link
+      data-cy="hackpack-link"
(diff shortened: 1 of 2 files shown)


src/components/user/resources/Hackpack.tsx
```diff
@@ -14,31 +14,34 @@ const Hackpack = ({ text, techs, link, description }: props) => {
...
+            >
+              <div className="text-hackathon-blue-100" data-cy="hackpack-icon">
+                {TECHSTACKS[tech]}
+              </div>
+              {tech}
+            </div>
+          ))}
+        </div>
+      </Link>
       <p className="mt-2">{description}</p>
```
shahdivyank, line 43: the entire card should be in the link component, so the description should also be included in the link

src/components/user/resources/Hackpack.tsx
```diff
@@ -14,32 +14,35 @@ const Hackpack = ({ text, techs, link, description }: props) => {
     <div className="rounded-xl bg-white p-3">
```
shahdivyank: remove the outer div, since there's only one child in the div, u can combine the properties of the div and the link tag

shahdivyank (approved):

PR #1853, opened 2024-09-18
removed default onChange and put it in parent

Shadow709:
image

### DIFF
--- a/src/components/Input.jsx
+++ b/src/components/Input.jsx
@@ -5,7 +5,7 @@ const Input = ({
   placeholder,
   value,
   editable = true,
-  onChange,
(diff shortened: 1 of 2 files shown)


src/components/Input.jsx
```diff
@@ -24,7 +23,7 @@ const Input = ({
         placeholder={placeholder}
         value={value}
         maxLength={maxLength}
-        onChange={(e) => setUser({ ...user, [name]: e.target.value })}
```
shahdivyank, line 27: needs to be tested, since the function now needs to be defined in the parents

Shadow-The-Dark-Knight:
image

PR #1955, opened 2025-02-17
clock edit and sizing

qhgill:
image


src/components/admin/services/timer/clock.jsx
```diff
@@ -91,15 +101,12 @@ const Timer = ({ onRemove }) => {
       <div className="mt-4">
         {play && (
           <Pause
-            onClick={() => !edit && setPlay(false)}
+            onClick={() => pressPause()}
```
shahdivyank: `onClick={pressPause}`

src/components/admin/services/timer/clock.jsx
```diff
@@ -91,15 +101,12 @@ const Timer = ({ onRemove }) => {
...
             className="hover:cursor-pointer"
           />
         )}
         {!play && (
-          <Play
-            onClick={() => !edit && setPlay(true)}
-            className="hover:cursor-pointer"
-          />
+          <Play onClick={() => pressPlay()} className="hover:cursor-pointer" />
```
shahdivyank: `onClick={pressPlay}`

src/components/admin/services/timer/controls.tsx
```diff
@@ -2,26 +2,20 @@ import { Trash2, Pen, Check } from "lucide-react";
...
 type props = {
   edit: boolean;
-  setEdit: (value: boolean) => void;
   onRemove: () => void;
 };
 
-const Controls = ({ edit, setEdit, onRemove }: props) => {
+const Controls = ({ edit, onRemove }: props) => {
   return (
     <div className="flex gap-4">
```
shahdivyank: what's the point of having the edit? it seems like no matter the condition it will render the same thing? so we dont need to pass in the edit variable here

src/components/admin/services/timer/controls.tsx
```diff
@@ -1,30 +1,13 @@
...
+import { Trash2 } from "lucide-react";
 
 type props = {
-  edit: boolean;
-  setEdit: (value: boolean) => void;
   onRemove: () => void;
 };
 
-const Controls = ({ edit, setEdit, onRemove }: props) => {
+const Controls = ({ onRemove }: props) => {
```
shahdivyank: dont need a component just to call a single div / icon, lets combine into the parent component

src/components/admin/services/timer/clock.jsx
```diff
@@ -49,7 +59,7 @@ const Timer = ({ onRemove }) => {
           className="flex-grow bg-transparent pl-2 text-3xl font-semibold outline-none"
           placeholder="Untitled Timer"
         />
-        <Controls edit={edit} setEdit={setEdit} onRemove={onRemove} />
+        <Controls edit={edit} onRemove={onRemove} />
```
shahdivyank: edit shouldn't be passed in anymore

src/components/admin/services/timer/clock.jsx
```diff
@@ -33,6 +33,16 @@ const Timer = ({ onRemove }) => {
     return () => clearInterval(id);
   }, [play, total]);
 
+  const pressPlay = () => {
+    setEdit(false);
```
shahdivyank: whats the edit currently being used for?

qhgill: makes timer non-editable while playing, can remove if needed but editing while the timer is going would be kinda jank

shahdivyank (approved):

PR #1956, opened 2025-02-24
Countdown Landing

seanquiambao:
image

Abstracted countdown digits into its own component (./clock.tsx).

image

can change colors with digits, and units color. im assuming i leave the green bg up to the developers.
(background - the block behind digit)


src/components/live/countdown.tsx
```diff
@@ -0,0 +1,18 @@
+import Clock from "@/components/ui/countdown";
+
+const Countdown = () => {
+  return (
+    <div className="inline-flex-col inline-block w-full rounded-lg bg-hackathon-green-300 text-center shadow-xl">
+      <div className="m-2 mb-0 font-bold text-white">HACKING ENDS IN</div>
```
shahdivyank: this file should not be necessary should just be called within the landing file

src/components/ui/countdown.tsx
```diff
@@ -0,0 +1,78 @@
+"use client";
+import { useState, useEffect } from "react";
+import data from "@/data/config";
+
+type CountdownClassNames = {
```
shahdivyank: should be able to combine into the type below

src/components/ui/countdown.tsx
```diff
@@ -0,0 +1,82 @@
...
+  unit: string;
+  classNames: {
+    unit: string;
+    digit: string;
+    background: string;
+  };
+}
+
+const Digits = ({ value, unit, classNames }: digitProps) => {
+  console.log(classNames);
```
shahdivyank: remove console log

src/components/live/index.tsx
```diff
@@ -20,6 +21,13 @@ const Live = () => {
       <Team />
       <Committees />
       <Judges />
+      <Countdown
```
shahdivyank: call it within the landing, since it will be integrated with other things within the landing

shahdivyank (approved):

PR #2001, opened 2025-05-01
Create user.ts

### LINKED ISSUE #1999: Abstract Schema Validations
Each role has some shared schemas for validation, but we should combine them into a user.ts file which has the shared schemas like firstName, lastName, email and what not and then we can import them into the other user types

Image

pranavgowrish:
close #1999

### DIFF
--- a/src/schemas/user.ts
+++ b/src/schemas/user.ts
@@ -1,0 +1,31 @@
+import { z } from "zod";
+import { AGES, SHIRTS, GENDERS, DIETS } from "@/data/form/information"; //Common imports found in 8 out of 12 schemas
+
+//following exports are common in 9 out of 12 schemas
+export const schema = z.object({
+  firstName: z.string().min(1, { message: "First name is required" }),
+  lastName: z.string().min(1, { message: "Last name is required" }),
+  email: z.string().email({ message: "Invalid email address" }),
+  discord: z.string().min(1, { message: "Discord username is invalid" }),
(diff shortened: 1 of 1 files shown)


src/schemas/user.ts
```diff
@@ -0,0 +1,38 @@
+import { z } from "zod";
+import {
+  GENDERS,
+  SHIRTS,
+  GRADES,
+  MAJORS,
+  AGES,
+  DIETS,
+} from "@/data/form/information"; // Common imports found in most schemas
```
khandrew1: remove comments

src/schemas/user.ts
```diff
@@ -0,0 +1,38 @@
...
+
+// following exports are common in 9 out of 12 schemas
```
khandrew1: what about the other 3? what are they and what are you doing to handle them.

khandrew1 (changes requested):

src/schemas/user.ts
pranavgowrish: This user.ts can be imported in the following schemas for common fields:

1. admin.ts
2.  committee.ts
3. judge.ts
4. lead.ts
5. mentor.ts
6. participant.ts
7. volunteer.ts

For the rest, the fields are few so there's little common:

1. feedback.ts
2. idea.ts
3. interest.ts
4. panel.ts
5. sponsor.ts

shahdivyank:
once the user.ts file is made, the other forms should be inheriting from it as well

